PR #2, opened 2018-07-10
Replace Googler aliases with their GitHub users.

philwo:
FYI @meisterT @laszlocsomor @buchgr

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,69 +1,71 @@
 # Bazel Proposals and Design Documents
 
-This is an index of all design documents and proposals for [Bazel](https://bazel.build).
+This is an index of all design documents and proposals for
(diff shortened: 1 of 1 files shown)


philwo:
lindleyf is not in our username <=> GitHub map, so I left the name as is.

laszlocsomor:
Thanks! WDYT about changing the user names to links to the github profiles?

philwo:
Arghh. I thought GitHub would do that automatically for "@" prefixed usernames...

Yeah, would definitely look nicer. I'll have a look.

philwo:
@laszlocsomor Done!

laurentlb:
lgtm

jin:
Much better idea :-)

PR #3, opened 2018-07-10
Add the leading and ending pipes for table markdown formatting

jin:
This re-adds the leading and ending pipes for table markdown formatting. Without the pipes, editors are unable to detect that the text chunk as a markdown table to apply automatic formatting, even though Github renders it properly.

Preview: https://github.com/bazelbuild/proposals/blob/f19f4cd0d975233c5dc5c514c4946ce7890f81fc/README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,69 +11,68 @@ list.
 
 ### Implemented
 
-Last updated | Title                                                                                                                                      | Author(s) alias                                  | Category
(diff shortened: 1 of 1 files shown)


philwo:
Mhm, our `mdformat` tool removed them, I guess. Sorry!

Do you know a different auto-formatter for GitHub Markdown that I could use in the future? :)

jin:
Ah. I personally use `markdown-table-align` in Emacs/markdown-mode for inline aligning. Not sure what's the best solution for CLI formatting.

PR #7, opened 2018-07-13
Add for review: design on computing the hash of a source directory

aehlig:
Add a document for review that details which information we do and do
not include in the hash of the output tree generated by a Skylark
external repository rule.

PR #9, opened 2018-07-13
Fix spelling error in Bazel Hashing Doc

petroseskinder:
Small nit in title of #7


petroseskinder:
reviewers: @jin @laurentlb

PR #11, opened 2018-07-16
Add template for design documents

laurentlb:


docs/template.md
```diff
@@ -0,0 +1,35 @@
...
+This section gives a short summary of the proposal.
+
+
+# Background
+
+This section can give context. For example, it can explain the current state and
+have pointers to previous design documents.
+
+
+# Proposal
```
jin: Should this be more granular? Objectives, Design, Alternatives Considered?

docs/template.md
```diff
@@ -0,0 +1,35 @@
+---
+created: 2018-07-16
```
jin: also add updated?

laurentlb: done

docs/template.md
laurentlb: I'm not sure. I suggest leaving more freedom to the writer, since it may be used for very different documents (we can update the template later if needed).

jin: sgtm

PR #16, opened 2018-07-24
windows,tests: design Bash-less test execution

laszlocsomor:
Tracking bug: https://github.com/bazelbuild/bazel/issues/5508

### DIFF
--- a/designs/2018-07-18-windows-native-test-runner.md
+++ b/designs/2018-07-18-windows-native-test-runner.md
@@ -1,0 +1,423 @@
+---
+created: 2018-07-18
+last updated: 2018-07-24
+status: To be reviewed
+reviewers: ulfjack (lead), lberki, dslomov
+title: Test execution on Windows without Bash
+tracking issues: "#5508, #4691, #4319"
+author: laszlocsomor
+---
+
+# Abstract
+
+Let's change how Bazel runs tests on Windows to no longer require Bash.
+
+# Background
+
+For every test, Bazel requires a complex Bash script to set up the environment
+and to run the test. Therefore Windows users need to install MSYS2 Bash to run
+tests. This is undesirable (see
+[issue #4319](https://github.com/bazelbuild/bazel/issues/4319)).
+
+Bazel should run tests without requiring Bash (see
+[issue #5508](https://github.com/bazelbuild/bazel/issues/5508)). This document
+explains how.
+
+# Current design
+
+## Test execution
+
+Bazel runs tests by executing `TestRunnerAction`s. Test actions are similar to
+build actions (e.g. `SpawnAction`): they take a list of input files, execute a
+command, and produce some output files.
+
+The inputs of the test action are the test binary and its dependencies, plus
+some extra helper files such as the test wrapper script
+(`@bazel_tools//tools/test/test-setup.sh`) and optionally the coverage collector
+and LCOV merger tools.
+
+The command of the test action is the test wrapper script, plus additional
+user-specified arguments. This script initializes the environment for the actual
+test binary, then runs the test.
+
+The outputs of the test action are the XML test log and the "undeclared outputs"
+file. The XML test log is an XML file that records two things: metadata about
+the test (such as the test target name, whether the test passed or failed), and
+the textual text log (that is, the output to `stdout` and `stderr`). The
+undeclared outputs file is a zip archive of files that the test created and are
+potentially interesting to the user.
+
+## Test termination
+
+Tests may terminate:
+
+*   cleanly, if the test process terminates by itself (regardless of whether the
+    test passed or failed)
+
+*   abruptly, if the test process is killed:
+
+    *   by the user, interrupting test execution using Ctrl+C
+
+    *   by Bazel, when a test times out or another test fails and
+        `--test_keep_going` is disabled).
+
+In every case the test action produces an XML test log.
+
+Bazel on Windows terminates tests abruptly in the following locations:
+
+*   [`WindowsSubprocesses.terminate()` in `WindowsSubprocess.waiterThreadFunc()`](https://github.com/bazelbuild/bazel/blob/a9c71b4c8a6d8b998c950c35c1e0a697a0d68ffa/src/main/java/com/google/devtools/build/lib/windows/WindowsSubprocess.java#L159),
+    when the test times out, and
+
+*   [`WindowsSubprocess.destroy()` in `FutureCommandResultImpl.waitForProcess()`](https://github.com/bazelbuild/bazel/blob/a9c71b4c8a6d8b998c950c35c1e0a697a0d68ffa/src/main/java/com/google/devtools/build/lib/shell/FutureCommandResultImpl.java#L99),
+    when the test is interrupted.
+
+## Test wrapper control flow
+
+1.  Absolutizes path-storing environment variables.
+
+    **Why**: At the time of creating the `TestRunnerAction` (along with its
+    environment), Bazel doesn't yet know the execution root the test will run
+    under. `test-setup.sh` absolutizes the envvars by making them relative to
+    `$PWD`.
+
+1.  Creates some directories, e.g. for the undeclared outputs, the shard status
+    file, the XML test log, and the test temp directory.
+
+1.  Exports some environment variables.
+
+    **Why**: Tests and test runners require envvars such as `$TEST_TMPDIR` and
+    `$TEST_SHARD_INDEX`.
+
+1.  Defines `rlocation()` to look up paths of data-dependencies.
+
+    **Why**:
+
+    *   `test-setup.sh` itself looks up the test executable's path.
+
+    *   For sake of shell tests (in case the actual test is a `sh_test`). This
+        use-case does not apply for the subset of shell tests that use the Bash
+        runfiles library in `@bazel_tools//tools/bash/runfiles`.
+
+1.  Defines `encode_output_file()`.
+
+    This function runs `perl` and `sed` to sanitize the textual test log for the
+    test XML test log's CDATA section. `write_xml_output_file()` calls this
+    function.
+
+1.  Defines `write_xml_output_file()` that:
+
+    *   Creates the test XML file, with the help of `encode_output_file()`.
+
+    *   Removes `${XML_OUTPUT_FILE}.log`.
+
+        **Why**: `${XML_OUTPUT_FILE}.log` is a temporary file containing the
+        test's raw output.
+
+1.  Changes the current directory to the test's runfiles directory (only when
+    coverage collection is disabled).
+
+    **Why**: Actions run in the execroot by default. Changing the directory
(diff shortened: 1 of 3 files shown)


laszlocsomor:
Lead reviewer: @ulfjack 
Optional reviewers: @lberki , @dslomov

laszlocsomor:
Thanks! To be clear, are you approving of the design itself, or just approving that the doc contains all important points and it's ready for review?

dslomov:
Ready for review, as per https://bazel.build/designs/index.html#create-a-pull-request

designs/2018-07-18-windows-native-test-runner.md
```diff
@@ -0,0 +1,423 @@
...
+    is in the current directory.
+
+    In fact this step is unnecessary, because the test executable's path is
+    always absolute.
+
+1.  Sets `$TEST_PATH` to the absolute path of the test executable.
+
+    If `$TEST_SHORT_EXEC_PATH` is defined, it sets an alternative `$TEST_PATH`.
+    
+    **Why**: To avoid too long paths on Windows with remote execution.
```
jasharpe: FYI: Microsoft has fixed the bug that required this so we should be able to drop this once the next Windows Server release is out (a few months from now likely).

laszlocsomor: Thanks! What exactly did they fix: is it now possible to set the CWD to a path longer than MAX_PATH?
Btw I think using a junction ("Design adequacy" > "step 9") will beat the purpose of  `$TEST_SHORT_EXEC_PATH`.

designs/2018-07-18-windows-native-test-runner.md
```diff
@@ -0,0 +1,423 @@
...
+
+## Guarded by a flag
+
+The new solution must be guarded by a flag. This allows us to roll out the
+feature in multiple stages, and users to easily revert to the old behavior in
+case the new one is buggy.
+
+# Non-requirements of the solution
+
+The solution does not have to cover remote test execution, because Bazel does
```
jasharpe: If I understand correctly, this isn't true. We run test-setup.sh on remote Windows machines now, and would likely run whatever wrapper binary replaces test-setup.sh on remote Windows machines as well.

That said, I don't see a reason why your proposal (which if I understand correctly is basically replacing test-setup.sh with a native program) wouldn't work with remote execution.

designs/2018-07-18-windows-native-test-runner.md
```diff
@@ -0,0 +1,423 @@
...
+## `rlocation()` support for `sh_test` rules
+
+The new test execution mechanism will not define `rlocation()` for the benefit
+of `sh_test` rules. See the [Backward compatibility](#backward-compatibility)
+section for details.
+
+# Implementation language
+
+We'll implement the test wrapper in C++, compile it as a x86\_64 Windows binary,
+and bundle it with Bazel as `@bazel_tools//tools/test:test-wrapper.exe`.
```
jasharpe: Out of curiosity, why bundle the binary rather than build it on the fly like the launchers are?

laszlocsomor: Because we want to require no C++ compiler. The launchers are actually pre-built: Bazel just appends data at the end of a pre-built binary.

designs/2018-07-18-windows-native-test-runner.md
laszlocsomor: Ah I see. But it seems, according to MSDN, SetCurrentDirectoryW now also supports long paths.

designs/2018-07-18-windows-native-test-runner.md
```diff
@@ -0,0 +1,423 @@
...
+## Guarded by a flag
+
+The new solution must be guarded by a flag. This allows us to roll out the
+feature in multiple stages, and users to easily revert to the old behavior in
+case the new one is buggy.
+
+# Non-requirements of the solution
+
+The solution does not have to cover remote test execution, because Bazel does
+not manage remote processes. To run a test remotely, Bazel sends a request to a
```
ulfjack: Bazel does upload the test-setup.sh script to the remote machine right now.

designs/2018-07-18-windows-native-test-runner.md
```diff
@@ -0,0 +1,423 @@
...
+## Processes
+
+There will be two or three processes per test (same numbers as today):
+
+*   a parent process, which runs the test wrapper
+
+*   a child process, which runs either the test binary or the coverage collector
+
+*   optionally a second child process after the first one finished, which runs
+    the `$LCOV_MERGER` when coverage collection is requested
```
ulfjack: We only run LCOV_MERGER if the test exits successfully.

designs/2018-07-18-windows-native-test-runner.md
```diff
@@ -0,0 +1,423 @@
...
+
+We change Bazel not to forcefully terminate the test wrapper process upon
+interruption, but instead:
+
+1.  ask the process to interrupt and shut down, by sending it a control message
+    on some channel
+
+1.  wait some time for the process to complete its shutdown protocol
+
+1.  forcefully terminate the process only if it's still running after a timeout.
```
ulfjack: Add explanation why you want to change that.

designs/2018-07-18-windows-native-test-runner.md
```diff
@@ -0,0 +1,423 @@
...
+
+The communication channel between Bazel and the test wrapper process will be the
+test wrapper's `stdin`.
+
+Using `stdin` is simple and convenient: the only supported control message is
+the request for interruption. For now a single byte will suffice as this
+message. This communication protocol is easily exendable if necessary.
+
+Using `stdin` is also safe: no other process has a handle to the test wrapper's
+`stdin`, so no other process will inadvertently send the interruption request.
```
ulfjack: We need to sync with any remote execution implementations on this.

designs/2018-07-18-windows-native-test-runner.md
```diff
@@ -0,0 +1,423 @@
...
+
+### Shutdown protocol
+
+When requested to interrupt and shut down, the test wrapper should exit as soon
+as possible.
+
+The primary output of test execution is the XML test log: it carries the most
+useful information for the user. The XML file records the test's status (passed
+or failed) and the test's textual output. The test wrapper should ensure it
```
ulfjack: I'm in the process of making Bazel ensure that a test.xml exists, rather than relying on the test wrapper.

designs/2018-07-18-windows-native-test-runner.md
```diff
@@ -0,0 +1,423 @@
...
+test execution mechanism. We will roll it out in several stages (see
+[rollout plan](#rollout-plan)) so users will have time to test it, report bugs,
+and revert to the old mechanism in case they discover bugs.
+
+# Rollout plan
+
+We will roll out this feature over several Bazel minor versions:
+
+1.  version `0.N.*`: contains both the new and old test execution mechanisms and
+    supports the `--[no]windows_bashless_test` flag. By default the flag is
```
ulfjack: Double negation is not ideal.

designs/2018-07-18-windows-native-test-runner.md
```diff
@@ -0,0 +1,423 @@
...
+
+Windows doesn't support process replacement (`exec(3)` on Unixes), therefore in
+order to set up the test's environment, the test setup process must create the
+test process.
+
+## Processes
+
+There will be two or three processes per test (same numbers as today):
+
+*   a parent process, which runs the test wrapper
```
lberki: Have you considered implementing this not as a separate executable, but as JNI code within the Bazel binary? This would bring a few advantages:

* One less process to create (process creation is expensive on Windows)
* No need to come up with a protocol between the Bazel process and the test wrapper
* One less moving part (we already have JNI code for process management)

The downside is that a badly-written test wrapper can bring down the whole Bazel process, but I assume the code wouldn't be that complicated for that to be a great risk.

designs/2018-07-18-windows-native-test-runner.md
```diff
@@ -0,0 +1,423 @@
...
+1.  ask the process to interrupt and shut down, by sending it a control message
+    on some channel
+
+1.  wait some time for the process to complete its shutdown protocol
+
+1.  forcefully terminate the process only if it's still running after a timeout.
+
+### Interruption request
+
+The communication channel between Bazel and the test wrapper process will be the
```
lberki: Is this the standard way to communicate with a child process on Windows?

designs/2018-07-18-windows-native-test-runner.md
```diff
@@ -0,0 +1,423 @@
...
+terminates the test process (see [Test termination](#test-termination)), leaving
+no chance for cleanup. Therefore in order to capture the textual test log and
+convert it to XML even when the test is interrupted, the test and the log
+capturer cannot run in the same process.
+
+Windows doesn't support process replacement (`exec(3)` on Unixes), therefore in
+order to set up the test's environment, the test setup process must create the
+test process.
+
+## Processes
```
lberki: Are these processes going to be wrapped in a Job object (like all the current subprocesses on Windows)?

designs/2018-07-18-windows-native-test-runner.md
laszlocsomor: Right, @jasharpe already mentioned that. I'll rework that section and ping @jasharpe 's comment thread.

dslomov:
Meta-points:
* per  https://bazel.build/designs/index.html#create-a-pull-request, review discussion needs to happen on bazel-dev thread, not on this pull request (which is just to get the document in the repo)
* review discussion here is quite difficult to follow :(
Markdown format is very bad at supporting the discussion, Google Docs are just so much better
I think the right way would be to file github issues for anything that needs clarification

laszlocsomor:
Re: https://github.com/bazelbuild/proposals/pull/16#issuecomment-407690267

> - per https://bazel.build/designs/index.html#create-a-pull-request, review discussion needs to happen on bazel-dev thread, not on this pull request (which is just to get the document in the repo)

I hear you, but I think the process is broken. The reality is, for everything else we do on GitHub (PRs, issues) we are used to commenting here. Expecting that people will behave differently just for proposal PRs is futile.

> - review discussion here is quite difficult to follow :(

I agree with that. That's a problem with GitHub's code review tool, IMO.

Perhaps MarkDown isn't the right format for design discussions and we should stick to GDocs, and maybe only use MarkDown for archival purposes. WDYT?

laszlocsomor:
Thanks everyone for reviews!

Let's continue the review comments on this email thread: https://groups.google.com/d/msg/bazel-dev/Vnugr-wDTqU/0tKcYkkyBwAJ

That email also explains the reasons. I'll copy all comment threads from this PR to that email thread.

laszlocsomor:
Actually, instead of trying to adjust this PR, let me create a new one just for the changes to `README.md`.

PR #17, opened 2018-07-24
Toolchain Modes

katre:
Tracking issue: https://github.com/bazelbuild/bazel/issues/3858

### DIFF
--- a/designs/2018-07-23-toolchain-modes.md
+++ b/designs/2018-07-23-toolchain-modes.md
@@ -1,0 +1,169 @@
+---
+created: 2018-07-23
+last updated: 2018-07-23
+status: To be reviewed
+reviewers:
+  - gregce
+title: Toolchain Modes
+authors:
+  - katre
+---
+
+# Background
+
+The current implementation of [toolchain
+resolution](https://docs.bazel.build/versions/master/toolchains.html#toolchain-resolution)
+is very simplistic: it assumes that there should only ever be one toolchain per
+target platform. However, that is not true: there are several reasons why there
+might be multiple valid toolchains for a given target platform:
+* Generating different types of output: i.e. optimized vs. debug outputs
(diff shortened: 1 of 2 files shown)


katre:
Lead reviewer: @gregestren

gregestren:
Review acknowledged.

katre:
@laurentlb, it looks like @laszlocsomor is keeping the draft design doc in his own fork of the proposals repository. Is that an approach I should try as well?

laurentlb:
From my point of view, both are okay (we're still trying to figure out the best workflow).

Adding the draft here means you need a pull request + approval for each change. The reviewer can point out typos, formatting or trivial issues. But GitHub pull requests are not well suited for the discussions. Since it was getting messy, @laszlocsomor moved his proposal to his GitHub, to keep all discussions in the mailing-list.

laszlocsomor:
@katre : I'm happy to share my experience so far if you're interested.

designs/2018-07-23-toolchain-modes.md
```diff
@@ -0,0 +1,169 @@
...
+  - katre
+---
+
+# Background
+
+The current implementation of [toolchain
+resolution](https://docs.bazel.build/versions/master/toolchains.html#toolchain-resolution)
+is very simplistic: it assumes that there should only ever be one toolchain per
+target platform. However, that is not true: there are several reasons why there
+might be multiple valid toolchains for a given target platform:
```
gregestren: While I (think) I get what this is saying, I'm also a bit confused. Doesn't the current logic grant that different toolchains can match the platform but just happens to choose the first match?

If "assumes there should only be one toolchain per platform" means assume only one is *selected*, does that really change with this proposal? i.e. if you build `bazel build //foo --compilation_mode`, you can tweak the compilation mode for the build but that setting still applies to the entire target platform.

PR #31, opened 2018-08-27
Mark "Args Directory Support" as implemented.

tomlu:


brandjon:
Approved. Note that in the future, this kind of change probably belongs to the [Starlark-specific proposal review process](https://github.com/bazelbuild/starlark/tree/master/proposals). In particular, that would require that the doc be markdown-ified, at least by the time it's finalized.

PR #35, opened 2018-10-02
Mark repository hashing as implemented

aehlig:
Repository hashing as described in the respective document is implemented
in bazel at HEAD and will be part of the 0.19 release, still guarded behind
an experimental flag. See also https://blog.bazel.build/2018/09/28/first-class-resolved-file.html


aiuto:
There was a backlog of "add document to the index" changes which I am approving because they should have been done or withdrawn long ago. The automerge failed on this. Can you take a look and resolve it.

PR #48, opened 2018-10-19
Mark Platform Inheritance approved.

katre:
@gregestren, you are the final approver, let me know if you have any further points or if I missed anything from the [design discussion thread](https://groups.google.com/forum/#!msg/bazel-dev/wKacIE4MIRM/hKWCYBF1BgAJ).


gregestren:
Ack. Reading through, and enthusiastically agree with the thread consensus of using 'parents" + single inheritance, since the case for multiple inheritance now isn't there.

designs/2018-10-09-platform-inheritance.md
```diff
@@ -1,7 +1,7 @@
 ---
 created: 2018-10-09
 last updated: 2018-10-09
```
gregestren: Update date?

README.md
```diff
@@ -44,6 +44,7 @@ list.
 
 | Last updated | Title                                                                                                                  | Author(s) alias                                                                    | Category              |
 | ------------ | ------------------------------------------------------------------------------------------------------------------     | ---------------------------------------------------------------------------------- | --------              |
+|   2018-10-10 | [Platform Inheritance](designs/2018-10-09-platform-inheritance.md)                                                     | [@katre](https://github.com/katre)       | Configurability |
```
gregestren: Update date here too?

This seems unfortunate from a process perspective (not your PR's fault of course), this manual reshuffling we always have to do whenever we update any doc. I wonder if a v2 design repository can automatically render these tables. That plus the ability to filter or organize by subject would be AWESOME.

designs/2018-10-09-platform-inheritance.md
```diff
@@ -1,7 +1,7 @@
...
-status: In Review
+status: Approved
```
gregestren: Nit: for "allow for later migration to allowing multiple inheritance" below: can we clarify that this just gives us API flexibility  vs. being an actual eventual commitment to multiple inheritance?

gregestren: Nit 2: final line mentions "parent attribute" instead of "parents attribute".

designs/2018-10-09-platform-inheritance.md
katre: Done.

README.md
katre: Yes, generating these automatically from the actual files would be nice (with placeholders for the non-markdown versions). Something for the future.

Date updated.

designs/2018-10-09-platform-inheritance.md
katre: Fixed both.

PR #56, opened 2018-10-29
Add a proposal to report progress of external repositories

aehlig:
It is a long-standing feature request to be able to report progress
about fetching external repositories. As currently, there is way for
Starlark repository rules to do so, this document proposes an extension
of the external-repositories build API to allow reporting of progress.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,6 +60,7 @@ list.
 
 | Last updated | Title                                                                                                                                | Author(s) alias                                                              | Category              |
 | ------------ | ------------------------------------------------------------------------------------------------------- | ------------------------------------------ | -------- |
+|   2018-10-29 | [Progress reporting for external repositories](designs/2018-10-29-external-repositories-reporting.md)                                | [@aehlig](https://github.com/aehlig)                                         | External Repositoires |
 |   2018-10-29 | [Multiplex persistent worker](https://docs.google.com/document/d/1OC0cVj1Y1QYo6n-wlK6mIwwG7xS2BJX7fuvf1r5LtnU/edit?usp=sharing)               | [@borkaehw](https://github.com/borkaehw) | Bazel |
 |   2018-10-24 | [Resource auto-tuning via flags](https://docs.google.com/document/d/11N3SXtjPgGnlu0-cUU9gVegUMr5Wrjj_h8PCTGdQlB8/edit)               | [@jmmv](https://github.com/jmmv) | Performance |
 |   2018-10-23 | [Auto-configured Host Platform](designs/2018-10-22-autoconfigured-host-platform.md)                                           | [@katre](https://github.com/katre)       | Configurability |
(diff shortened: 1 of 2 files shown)


brandjon:
Nit: Reword abstract to give a tiny bit more context than just the API change. It's enough to say this adds a way for repository rules to emit progress messages to the console during WORKSPACE evaluation.

aehlig:
> I don't understand this sentence:
>
> > Instead of the the progress_message of an action, the repository
> > name together with the argument of the last call to report_progress
> > will be used for an ongoing fetch.

The context is given by the sentence immediately before it: the UI
is expected to present fetches (i.e., the execution of an external
repository rule) in a similar way as regular actions during the
build; the latter have a progress message and for the former, the
call to `report_progress` sets the corresponding value (and hence,
it might change during the execution of an external repository
rule).

> What is the API? I thought it accepted a single string parameter.

Yes. The `report_progress` functions takes a single string as
argument. That string will be the progress status for the fetch of
that external repository from the moment of call to the function
onwards, until set to a new value by another such funciton call
(or the finish of the fetch for the external repository).

> Typos: "statrted", "the the".

done

> What is the expected UI behavior regarding newlines? Repo rule
> authors need to know whether they have the option to include them,
> whether they're allowed but discouraged, or whether they'll be
> silently stripped.

Same as for actions. Made that more explicit.

-- 
Klaus Aehlig
Google Germany GmbH, Erika-Mann-Str. 33, 80636 Muenchen
Registergericht und -nummer: Hamburg, HRB 86891
Sitz der Gesellschaft: Hamburg
Geschaeftsfuehrer: Paul Terence Manicle, Halimah DeLaine Prado

brandjon:
Ok, I was missing that "will be used for" was referring only to the UI, not the API. LGTM for merge as draft (though I'm not a reviewer).

aehlig:
> LGTM for merging as draft as I believe that's the next step?

Yes, the next step is to send over the mailing list, that this proposal is under review and ask everybody for feedback&mdash;and in particular feedback from the reviewers.

> Can you add an example of what a repository rule that is using this function would look like?

Will add such an example as a prototype patch tomorrow.

aehlig:
> > LGTM for merging as draft as I believe that's the next step?
>
> Yes, the next step is to send over the mailing list, that this
> proposal is under review and ask everybody for feedback—and in
> particular feedback from the reviewers.

Discussion: https://groups.google.com/forum/#!topic/bazel-dev/cKIXXIYRoYo

> > Can you add an example of what a repository rule that is using
> > this function would look like?
>
> Will add such an example as a prototype patch tomorrow.

Example usage: https://bazel-review.googlesource.com/c/bazel/+/80370 

-- 
Klaus Aehlig
Google Germany GmbH, Erika-Mann-Str. 33, 80636 Muenchen
Registergericht und -nummer: Hamburg, HRB 86891
Sitz der Gesellschaft: Hamburg
Geschaeftsfuehrer: Paul Terence Manicle, Halimah DeLaine Prado

PR #65, opened 2018-11-09
Add proposal: Config Setting  Chaining

gregestren:
As per https://bazel.build/designs/, this is a request to add the proposal as a draft to the design doc registry.  This is not an actual design review, just a review that the basic structure fits with our design process.

After this PR I'll put out an announcement and start formal review.


gregestren:
Ah, here's precise instructions for this stage: https://bazel.build/designs/#when-you-receive-a-new-proposal

I'm not trying to assume you don't know this already, but I'm aware this is a new process for most of us and I'm still learning the ropes myself.

PR #81, opened 2018-12-14
Promote Versioned Documentation to Implemented

jin:
It is live. https://docs.bazel.build/versions/0.20.0/bazel-overview.html


brandjon:
Woohoo!

PR #82, opened 2019-01-14
Add a design on delayed load statements

aehlig:
Add a design document discussing the use cases of a `load` statement
in starlark that is evaluated only on first use of the loaded
symbol.


designs/2019-01-14-delayed-load.md
```diff
@@ -0,0 +1,90 @@
...
+the `deps` functions provided by the dependencies they add.
+
+## Suggested changes
+
+To support, in a `WORKSPACE` context, loading files from repositories
+that still have to be defined, we suggest adding a new boolean
+parameter `delay_load` to the `load` statement, defaulting to
+`False`. (As the value cannot be a string, there is no overlap
+with with aliased imports; but we can still decide to disallow an
+import to be aliased to `delay_load`.) It will be an error if a
```
dkelmer: Is there a value other than "True" or "False"? If not, how about: 

"It will be an error if `delay_load` is 'True' and the file containing..."

designs/2019-01-14-delayed-load.md
```diff
@@ -0,0 +1,90 @@
...
+value different from `False` is used if the file containing the
+`load` statement is not included (directly or indirectly) from the
+(top-level) `WOKRSPACE` file.
```
dkelmer: But the error would only be discovered when the bzl file actually gets loaded when the symbol is used, right? Can you clarify that in a subsequent sentence?

designs/2019-01-14-delayed-load.md
```diff
@@ -0,0 +1,90 @@
...
+
+If a `load` statement with `delay_load` set to `True` is executed,
+it binds the corresponding variables not to a value, but to a thunk
```
dkelmer: What is a "thunk"? :)

designs/2019-01-14-delayed-load.md
```diff
@@ -0,0 +1,90 @@
...
+that is evaluated on first use. On that first use, the repository
+with the file to be loaded has to be already defined, otherwise
+bazel will error out. At the same time, the definition of the
+external repository the file load from belongs to (if was taken
```
dkelmer: "the external repository that the file was loaded from (if it was taken from an external repository) will be made immutable"

dkelmer:
I assume you'll start a thread for this design, and I recommend linking to that thread from the recursive workspace design thread as well.

designs/2019-01-14-delayed-load.md
asuffield: https://en.wikipedia.org/wiki/Thunk (the term is a little obscure, dates from the 1960s)

designs/2019-01-14-delayed-load.md
```diff
@@ -0,0 +1,90 @@
...
+its dependencies. These `deps` functions can be loaded and called
+from a `WORKSPACE` file adding these repositories as dependencies.
+However, these `deps` function cannot load (and hence not call)
+the `deps` functions provided by the dependencies they add.
+
+## Suggested changes
+
+To support, in a `WORKSPACE` context, loading files from repositories
+that still have to be defined, we suggest adding a new boolean
+parameter `delay_load` to the `load` statement, defaulting to
```
asuffield: This is probably the part I find the most syntactically awkward, because there's already a meaning to kwargs in the load() function, so the semantics then become: "The name (y) specifies the local name, unless it is delay_load in which case it changes the behaviour of the statement..."

Is there anywhere we can sneak in non-ambiguous syntax? Or should we just create a delay_load() statement or similar?

designs/2019-01-14-delayed-load.md
asuffield: I agree with the concept, but I would ask whether external repository definitions should be made immutable as soon as they are first defined - I thought that was how it already worked.

designs/2019-01-14-delayed-load.md
```diff
@@ -0,0 +1,90 @@
...
+
+For the `WORKSPACE`, however, the situation is different. New
+external repositories are still being defined here, thus providing
+new locations where files could be loaded from. To deal with this
+special situation, the top-level `WORKSPACE` file is split at each
+block of top-level `load` statements. Each of the resulting chunks
+is interpreted separately with its own environment. This splitting,
+however, only happens in the `WORKSPACE` file and on top level,
+thus allowing for definitions directly in the `WORKSPACE` file that
+cannot be moved to a macro.
```
asuffield: Is this documented somewhere already? We should stash this block of text somewhere more permanent, it's quite illuminating.

designs/2019-01-14-delayed-load.md
asuffield: Also worth clarifying the case where the symbol is never used. At first I thought "they should all be evaluated at the end just to be sure", but then I realised there's scope for creative programming with symbols that are only used if they happen to be defined, of the form:

```load("@bar//:deps.bzl", "bar_deps", delay_load=True)
load("@baz//:deps.bzl", "baz_deps", delay_load=True)

def foo_deps()
  if foo:
    http_archive(name="bar", ...)
    bar_deps()
  else:
    http_archive(name="baz", ...)
    baz_deps
```

So I'm tempted to say that should work - if nothing calls it, the load statement is ignored.

aehlig: > But the error would only be discovered when the bzl file actually gets loaded when the symbol is used, right? Can you clarify that in a subsequent sentence?

An attempt to delay a load outside a workspace context is already an error. For a `BUILD` file context there is no need to delay loading, as the files are already defined. Explictly stated that this feature is only available in a workspace context.

> Also worth clarifying the case where the symbol is never used.

Explicitly stated the corollary of "evaluated at first use" that it will never be evaluated if the symbol is not used.

designs/2019-01-14-delayed-load.md
aehlig: > "the external repository that the file was loaded from (if it was taken from an external repository) will be made immutable"

done

aehlig: > I agree with the concept, but I would ask whether external repository definitions should be made immutable as soon as they are first defined - I thought that was how it already worked.

The current semantics of a `WORKSAPCE` file is that the latest definition before the first `load` following the first definition is the one that wins. So the suggested semantics is in line with what currently happens in `WORKSPACE` files.

designs/2019-01-14-delayed-load.md
aehlig: The type of the argument (String versus boolean) already distinguishes the cases. But calling the function `delayed_load` would also work.

PR #94, opened 2019-02-12
Add design drafts for execution and toolchain transitions.

katre:


designs/2019-02-12-execution-transitions.md
```diff
@@ -0,0 +1,76 @@
...
+execution platform's label will be added to an internal-only flag on
+[PlatformOptions](https://source.bazel.build/bazel/+/master:src/main/java/com/google/devtools/build/lib/analysis/PlatformOptions.java).
+If no execution platform is selected, this internal-only flag will be cleared.
+Then, when the new execution transition is used, the following steps will take
+place:
+
+1.  The previous execution platform's label will be copied from the
+    internal-only flag to the `--platforms` flag.
+1.  The internal-only flag will be cleared.
+1.  For all fragments, the
```
brandjon: I think you want this step to happen before you set the `--platforms` flag or you'll overwrite it with the host platform.

designs/2019-02-12-execution-transitions.md
```diff
@@ -0,0 +1,76 @@
...
+
+In order to properly transition to a configuration where the previous execution
+platform is the current target platform, the execution platform must be
+available to the transition. Note that there is no flag to directly set the
+execution platform (the closest flag available is
+[`--extra_execution_platforms`](https://source.bazel.build/bazel/+/master:src/main/java/com/google/devtools/build/lib/analysis/PlatformOptions.java;l=56?q=extra_execution_platforms),
+which adds available execution platforms, but does not select one).
+
+To enable the new transition, during toolchain resolution, the selected
+execution platform's label will be added to an internal-only flag on
```
brandjon: > internal-only flag on PlatformOptions

Is this to say that the execution platform will become part of the configuration? Does this have caching concerns when two otherwise identical builds occur on different execution platforms? (Their exec dependencies would clearly have to be different, but maybe you could otherwise reuse the result of the build of one of them for the other?)

If it's internal-only, how can a user make use of this ability when defining their own SBC transitions? Would they be forced to compose their Starlark-defined transition with an exec transition, perhaps using macros as syntactic sugar?

katre:
Please note: There's going to be a separate review thread to discuss the content: this PR is just to submit the draft for discussion.

designs/2019-02-12-toolchain-transitions.md
```diff
@@ -0,0 +1,110 @@
...
+  - gregce
+title: Toolchain Transitions
+authors:
+  - katre
+---
+
+# Abstract
+
+Toolchains need access to two separate platforms, both relating to the target
+that requires the toolchain: the target platform of the original target, and the
```
gregestren: Small nit: when I first read this I saw three cases: "the target that requires the toolchain", "the target platform of the original target", and "the execution platform of the original target". And got a bit confused. On second reading I saw my mistake and see that that grammar is already correct, so it really was my mistake. So I'm not sure if there's any way to actually be clearer. But just FYI.

designs/2019-02-12-execution-transitions.md
```diff
@@ -0,0 +1,76 @@
...
+  - gregce
+title: Execution Transitions
+authors:
+  - katre
+---
+
+# Abstract
+
+Historically, Bazel has had a single configuration used for targets that need to
+be executed on the host. This is called the **host configuration**, and is created
```
gregestren: Can we link to other context more fully explaining these concepts (maybe the doc that first explained host vs. execution configurations, or just Bazel docs that mention this)? I get it, but I also remember needing some effort to absorb the differences when I first heard about the concepts. And especially since this proposal suggests mixing the concepts of host and execution (via getHost()), I could see a first reader being confused about what really makes them different.

Also, is there any story on multiple execution configs and/or exec_compatible_with?

designs/2019-02-12-execution-transitions.md
gregestren: I was also going to ask if there should be a discussion on caching consequences, whatever they are.

designs/2019-02-12-toolchain-transitions.md
```diff
@@ -0,0 +1,110 @@
...
+cleared during toolchain resolution for the toolchain itself. This may happen
+using a marker to prevent the clearing, or by using a second internal-only flag.
+Then, when the toolchain-&gt;tool dependency (`foo_toolchain_impl` -&gt; `compiler`)
+is analyzed, the execution transition can change the target platform to be the
+correct execution platform, using the previously defined execution transition.
+
+Secondly, the toolchain-&gt;library dependency (`foo_toolchain_impl` -&gt;
+`standard_library`) can use the identity transition, to inherit the original
+target's configuration. This avoids the need to transition to a special
+configuration, and then back to the original target configuration.
```
gregestren: I think I like this approach, and its implications! "Passing along" original target configurations after transitions can be a weird design wart if not handled elegantly.

designs/2019-02-12-toolchain-transitions.md
```diff
@@ -0,0 +1,110 @@
...
+    toolchain resolution). The configuration of this target is currently not
+    well defined.
+1.  `standard_library`: Because the `foo_library` rule intends to directly link
+    this library into the output of `example`, it also needs to be in the
+    initial configuration (C1).
+1.  `compiler` - Because the `foo_library` rule intends to use this as the
+    executable for an action while building the output of `example`, it needs to
+    be in a different configuration (C2). C2 is the result of using the
+    [Execution Transition](2019-02-12-execution-transitions.md) on C1 from
+    `example`.
```
gregestren: I wonder if this should be the default, in the belief that more attributes are exec-platform focused? Is that even an accurate belief?

gregestren: Does "on C1 from `example`" imply `example` sets the execution transition or just the transition is off C1? Disambiguate?

designs/2019-02-12-execution-transitions.md
katre: Added comments on caching.

Composing transitions is not currently supported from Starlark, I think. @gregestren, is this something being worked on in the api?

designs/2019-02-12-toolchain-transitions.md
katre: Updated the wording, hope this is clearer.

designs/2019-02-12-execution-transitions.md
katre: Linked to the docs on configuration.

There will be many execution configurations, that is unavoidable with multiple execution platforms. The `exec_compatible_with` attribute (I think that's what you mean?) influences the choice of execution platform, and therefore what the execution configuration looks like, but I'm not sure what additional story is needed.

designs/2019-02-12-toolchain-transitions.md
katre: It's not easy to figure out a way to change the "default" transition in a Starlark rule definition. One option suggested by @brandjon was for a new parameter to the rule function that declares what the default transition is, but this will still make migration difficult and be a source of confusion for rule authors.

katre: Reworded.

designs/2019-02-12-execution-transitions.md
gregestren: If the implementation is internal, I guess Bazel-internal logic would compose in whatever order makes sense? Is there really any complicated requirement on that?

designs/2019-02-12-toolchain-transitions.md
gregestren: LGTM.

designs/2019-02-12-execution-transitions.md
gregestren: Yeah, I'm just clarifying that that's a distinction between host and exec configs that makes the difference clearer.  It could be nice to reference that somewhere for anyone not already familiar with those connections.

designs/2019-02-12-execution-transitions.md
katre: Toolchains will need to declare which transition to use out of the toolchain, like this:
```
my_toolchain = rule(
  _impl,
  attrs = {
    'compiler': attr.label(cfg = 'exec'),
  },
)
```

There is no way that I know to compose the built-in "exec" transition with a custom Starlark transition (created with the new `transition()` function). It would be handy to be able to do so, but we don't have syntax for that yet, and I think it's out of scope for this proposal.

PR #96, opened 2019-02-25
Update toolchain transitions design

katre:
Reflects the latest comments from the design review thread.

Please comment only on formatting and style: actual design discussion should still take place at [the review thread](https://groups.google.com/forum/!msg/bazel-dev/5osWxhoF0Fk/1yw4MXrEBQAJ).

### DIFF
--- a/designs/2019-02-12-execution-transitions.md
+++ b/designs/2019-02-12-execution-transitions.md
@@ -41,12 +41,13 @@ If no execution platform is selected, this internal-only flag will be cleared.
 Then, when the new execution transition is used, the following steps will take
 place:
 
-1.  The previous execution platform's label will be copied from the
-    internal-only flag to the `--platforms` flag.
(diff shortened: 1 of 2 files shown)


brandjon:
> As an example, the `genrule` rule has an attribute named `tools`, which should be built in the context of the execution platform, instead of the current host platform.

Did you mean to say that `tools` is changing to use the execution platform, or was that supposed to say `exec_tools` and describe the addition of a new attribute? Or was it describing a hypothetical end-state after a long migration process?

katre:
@brandjon This is describing the eventual goal after the migration, which will take quite a long time.

gregestren:
FYI  the review thread link doesn't work for me.

katre:
I removed too many characters, the review thread is https://groups.google.com/forum/#!msg/bazel-dev/5osWxhoF0Fk/1yw4MXrEBQAJ

PR #97, opened 2019-02-27
2019-02-26_RemoteFileSystem-For-Bazel.md

rahul-nitkkr:
Proposal for introducing a remote file system for bazel which does away with the need of having source code on disk.

### DIFF
--- a/designs/2019-02-26_RemoteFileSystem-For-Bazel.md
+++ b/designs/2019-02-26_RemoteFileSystem-For-Bazel.md
@@ -2,7 +2,8 @@
 created: 2019-02-26
 last updated: 2019-02-26
 status: Draft
-reviewers:
+reviewers: 
+  - ulfjack
 title: Remote File System for Bazel
 authors:
   - royrah
--- a/designs/2019-02-26_RemoteFileSystem-For-Bazel.md
+++ b/designs/2019-02-26_RemoteFileSystem-For-Bazel.md
@@ -13,7 +13,7 @@ discussion thread:
 # Overview
 Bazel as a build tool natively expects files of the build action requested to be present on Disk. For a monorepo of a significant size, dumping out the entire repository for every single build , even with bazel's caching is going to be a significant overhead.
 
-Statistically, a repo which is ~2 GBs of source needs around 1-2 mins to dump out to disk. This proposal intends to address this particular issue by extending bazel to support a Remote File system , which can talk to any repository which supports Content addressable storage of source files to be able to provide on demand , requested files/contents hashes/file stats via HTTP/GRPC endpoints.
+Statistically, a repo which is ~2 GBs of source needs around 1-2 mins to dump out to disk. This proposal intends to address this particular issue by extending bazel to support a Remote File system , which can talk to any repository which supports Content addressable storage of source files to be able to provide on demand , requested files/contents hashes/file stats via HTTP/GRPC endpoints. Alternatively , a FUSE store can be specified as the type and associated mount point can be passed along as a startup option.
 
 # Existing Approaches
 Currently , the bazel server does not support OOTB any way to get around having files on the local disk where bazel server is running. Even with the remote [ BuildFarm ] setup , the local files are uploaded to the remote server for being actioned on.
@@ -53,41 +53,6 @@ RemoteFileSystem options should at the minimum be the following :
 4. **--remote_filesystem_prefixes** : enables the delegating file system to decide what paths should be delegated to the RemoteFileSystem
 5. **--remote_filesystem_version** : optional parameter to point to version of the underlying repository to be used , defaults to master
 
-```Java
-public static DelegatingFileSystem make(
-        FileSystem writeDelegate, FileSystem remoteDelegate, PathFragment ... customPrefixes ) throws DigestHashFunction.DefaultHashFunctionNotSetException {
-   //Basic checks to ensure read and write delegates are compliant, then call a private constructor
-
- return new DelegatingFileSystem(writeDelegate, readDelegate, customPrefixes);
-}
-private FileSystem getDelegate(Path path) {
-    //based of custom prefixes/other implementation detail figure out the correct handler for operations
-}
-```
-
-An example :
-```Java
-@Override
-protected long getFileSize(Path path, boolean followSymlinks) throws IOException {
-    return getDelegate(path).getFileSize(path, followSymlinks);
-}
-```
-The RemoteFileSystem consists of the client which makes the HTTP/GRPC calls to the backend server .Minimally looks as follows :
-
-```Java
-private RemoteFileSystemClient remoteClient;
-
-public RemoteFileSystem() throws DigestHashFunction.DefaultHashFunctionNotSetException {}
-
-public RemoteFileSystem(DigestHashFunction hashFunction) {
-    super(hashFunction);
-}
-
-public RemoteFileSystem(RemoteFileSystemOptions remoteOptions, PathFragment workspaceDir, BlazeServerStartupOptions startupOptions)
-        throws DigestHashFunction.DefaultHashFunctionNotSetException {
- remoteClient = RemoteFileSystemClientFactory.create(remoteOptions, workspaceDir, startupOptions);
-}
-```
 
 Any server attempting to support a remote file system implementation , will need to support the following requests :
 
@@ -118,14 +83,25 @@ import "build/bazel/remote/execution/v2/remote_execution.proto";
 //file system enabled.
 service RemoteFileSystem {
 
-    rpc FetchData(FileSystemRequest) returns (FileSystemResponse) {
-        option (google.api.http) = { get: "/remotefs/{type}/{repo_version}/{path}" };
+    rpc FetchDigest(FileSystemRequest) returns (build.bazel.remote.execution.v2.Digest) {
+        option (google.api.http) = { get: "/remotefs/digest/{repo_version}/{path}" };
+    }
+
+    rpc FetchStat(FileSystemRequest) returns (FileStat) {
+        option (google.api.http) = { get: "/remotefs/file_stat/{repo_version}/{path}" };
+    }
+
+    rpc FetchDirents(FileSystemRequest) returns (DirectoryEntries) {
+        option (google.api.http) = { get: "/remotefs/dirents/{repo_version}/{path}" };
+    }
+
+    rpc FetchContent(FileSystemRequest) returns (stream DataChunk) {
+        option (google.api.http) = { get: "/remotefs/file_content/{repo_version}/{path}" };
     }
 }
 
-//a request to the remote file system is almost exclusively always comprised of 3 components
-//the type of the request,  a path to resolve , and the version of the repository against which to find the path
-//The type parameter helps the server respond appropriately to different queries
+//a request to the remote file system is always comprised of 2 components
+//a path to resolve , and the version of the repository against which to find the path
 message FileSystemRequest {
     //the relative path for this request, note that this relative to the WORKSPACE
     string path = 1;
@@ -136,31 +112,6 @@ message FileSystemRequest {
     //default value for this be master , and can be overridden with an empty string to if the underlying
     //FileSystem of the remote server doesn't rely on versions.
     string repo_version = 2;
-
-    enum RequestType {
-        //request for hash of the file denoted by the path
-        DIGEST = 0;
-
-        //request for file state returning {@FileStat}
-        FILE_STAT = 1;
-
-        //request for listing directory entries at a particular path
-        DIRENTS = 2;
-
-        //request for the actual byte content of a file
-        FILE_CONTENT = 3;
-    }
-
-    RequestType type = 3;
-}
-
-message FileSystemResponse {
-    oneof response {
-        build.bazel.remote.execution.v2.Digest digest = 1;
-        DirectoryEntries entries = 2;
-        FileStat filestat = 3;
-        OutputFile filecontent = 4;
-    }
 }
 
 //collection containing the names of all entities within the directory denoted by the path
@@ -192,34 +143,28 @@ message FileStat {
     //like file sysytems
     int64 last_change_time = 4;
 
-    //Is the requested path a regular file ?
-    bool isFile = 5;
+    //Type of the file requested corresponding to the path
+    enum FileType {
+        //a regular file
+        REGULAR_FILE = 0;
+
+        // a Direcotry
+        DIRECTORY = 1;
+
+        //Symlink or jucntion ( for windows )
+        SYMLINK = 2;
+    }
+
+    FileType fileType = 5;
 }
 
-message OutputFile {
-    // The full path of the file relative to the input root, including the
-    // filename. The path separator is a forward slash `/`.
-    string path = 1;
-
-    // The digest of the file's content.
-    build.bazel.remote.execution.v2.Digest digest = 2;
-
-    // The raw content of the file.
-    //
-    // This field may be used by the server to provide the content of a file
-    // inline in an
-    // [ActionResult][google.devtools.remoteexecution.v1test.ActionResult] and
-    // avoid requiring that the client make a separate call to
-    // [ContentAddressableStorage.GetBlob] to retrieve it.
-    //
-    // The client SHOULD NOT assume that it will get raw content with any request,
-    // and always be prepared to retrieve it via `digest`.
-    bytes content = 3;
+message DataChunk {
+    // The raw content of the file, chunked to 1MB blocks.
+    bytes content = 1;
 }
 
 ```
 
 # Caveats
-1. This implementation currently, causes a server restart on change of the source tree ( branch in git terminology ). 
-2. There is presently a need to have BUILD/WORKSPACE files locally dumped to disk before kicking off the build, Need to figure out if atleast the BUILD files can be remoted too.
+1. There is presently a need to have BUILD/WORKSPACE files locally dumped to disk before kicking off the build.
 
--- a/designs/2019-02-26_RemoteFileSystem-For-Bazel.md
+++ b/designs/2019-02-26_RemoteFileSystem-For-Bazel.md
@@ -1,0 +1,224 @@
+---
+created: 2019-02-26
+last updated: 2019-02-26
+status: Draft
+reviewers:
+title: Remote File System for Bazel
+authors:
+  - royrah
+discussion thread: 
+---
+
+# Overview
+Bazel as a build tool natively expects files of the build action requested to be present on Disk. For a monorepo of a significant size, dumping out the entire repository for every single build , even with bazel's caching is going to be a significant overhead.
+
+Statistically, a repo which is ~2 GBs of source needs around 1-2 mins to dump out to disk. This proposal intends to address this particular issue by extending bazel to support a Remote File system , which can talk to any repository which supports Content addressable storage of source files to be able to provide on demand , requested files/contents hashes/file stats via HTTP/GRPC endpoints.
+
+# Existing Approaches
+Currently , the bazel server does not support OOTB any way to get around having files on the local disk where bazel server is running. Even with the remote [ BuildFarm ] setup , the local files are uploaded to the remote server for being actioned on.
+
+# Requirements
+The extended RemoteFileSystem needs to support at minimum the following requests which bazel makes to build it's action graph by reading files :
+
+### Directory listing of source Globs
+Every build target has a glob of input files which bazel resolves during the loading phase to build an action graph of a target. The Remote endpoint needs to resolve the requested glob from the underlying filesystem. This does not get the actual content , but only the list of files 
+in the directory.
+
+### Content Hashes of glob paths and dependency paths
+Bazel is fast because it caches the content hashes of every path it has built previously and just provides them when an Action requesting the particular path is requested. The remote endpoint needs to be able to provide a content hash[ Digest ] for a requested path. The digest is the identifier for a any file in the CAS when executing in remote mode.
+
+### Actual Files for a requested path
+When an action is actually executed the worker needs the actual file. The remote endpoint needs to resolve the path and send back the byte content of the file.
+
+### Resolve a particular path as onDisk/remote
+There are cases when a particular implementation will be a hybrid of a remote and local file system, such as cases where third part dependencies are actually present on disk and source files are not. The remote filesystem needs to delegate requests for file paths to the appropriate handling filesystem.
+
+This design currently needs the BUILD, WORKSPACE and .bzl files to be present on disk to be actually able to resolve the targets  to be built. A true-remote FS should have the ability to fetch these on demand from the remote endpoints and then proceed.
+
+Additionally, there should be a set of options to customize the remote behaviour based on the underlying store [ FUSE ] as well as the protocol to be supported [ HTTP/GRPC ].
+
+# Suggested Implementation
+We propose to have RemoteFileSystem which can talk to a HTTP/GRPC file server to satisfy the above listed requirements. The FileSystem at minimum needs to extend AbstractFileSystemWithCustomStat to intercept all Filesystem requests, and then appropiately delegate based on a list of either custom prefixes/any other delegation implementation.
+
+The BazelFileSystemModule needs to be extended to support the new Filesystem, and a set of Options need to be provided to provide the endpoint [ hosts / port ] as well as authentication options.
+
+RemoteFileSystem options should at the minimum be the following :
+
+### Startup Options
+
+1. **--enable_remote_filesystem** : flag to enable the intercation with remote file system, off by default
+2. **--remote_filesystem_type** : one of HTTP/GRPC...and FUSE ?
+3. **--remote_filesystem_host** : host:port for the backend server for the filesystem requests in case of HTTP/GRPC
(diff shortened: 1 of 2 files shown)


rahul-nitkkr:
> 1. Please link to this proposal  from the README.md
>> Updated the README as requested
> 2. Have you identified a lead reviewer for this yet? (see https://bazel.build/designs/) If not, I can try to find one for you.
>> I was think of having @ulfjack as the lead reviewer, since we have had a discussion on the same in bazel-dev. (see https://groups.google.com/forum/m/#!topic/bazel-dev/2sov79W41xk)

designs/2019-02-26_RemoteFileSystem-For-Bazel.md
```diff
@@ -0,0 +1,225 @@
...
+discussion thread: 
+---
+
+# Overview
+Bazel as a build tool natively expects files of the build action requested to be present on Disk. For a monorepo of a significant size, dumping out the entire repository for every single build , even with bazel's caching is going to be a significant overhead.
+
+Statistically, a repo which is ~2 GBs of source needs around 1-2 mins to dump out to disk. This proposal intends to address this particular issue by extending bazel to support a Remote File system , which can talk to any repository which supports Content addressable storage of source files to be able to provide on demand , requested files/contents hashes/file stats via HTTP/GRPC endpoints.
+
+# Existing Approaches
+Currently , the bazel server does not support OOTB any way to get around having files on the local disk where bazel server is running. Even with the remote [ BuildFarm ] setup , the local files are uploaded to the remote server for being actioned on.
```
ulfjack: This should talk about FUSE as an alternative.

designs/2019-02-26_RemoteFileSystem-For-Bazel.md
```diff
@@ -0,0 +1,225 @@
...
+### Startup Options
+
+1. **--enable_remote_filesystem** : flag to enable the intercation with remote file system, off by default
+2. **--remote_filesystem_type** : one of HTTP/GRPC...and FUSE ?
+3. **--remote_filesystem_host** : host:port for the backend server for the filesystem requests in case of HTTP/GRPC
+4. **--remote_filesystem_prefixes** : enables the delegating file system to decide what paths should be delegated to the RemoteFileSystem
+5. **--remote_filesystem_version** : optional parameter to point to version of the underlying repository to be used , defaults to master
+
+```Java
+public static DelegatingFileSystem make(
```
ulfjack: I think it would be better to talk about the HTTP/GRPC protocol design here rather than the Java file system interface.

ulfjack: I see you have that below - maybe just remove this part?

designs/2019-02-26_RemoteFileSystem-For-Bazel.md
```diff
@@ -0,0 +1,225 @@
...
+        //request for hash of the file denoted by the path
+        DIGEST = 0;
+
+        //request for file state returning {@FileStat}
+        FILE_STAT = 1;
+
+        //request for listing directory entries at a particular path
+        DIRENTS = 2;
+
+        //request for the actual byte content of a file
```
ulfjack: What does this do for symlinks or directories?

designs/2019-02-26_RemoteFileSystem-For-Bazel.md
```diff
@@ -0,0 +1,225 @@
...
+    string path = 1;
+
+    //version/commit of the repository against which the above path needs to be resolve.
+    //Remote filesystems will be usually backed by a repository like Git, which rely on a version
+    //to fetch a file/content from.
+    //default value for this be master , and can be overridden with an empty string to if the underlying
+    //FileSystem of the remote server doesn't rely on versions.
+    string repo_version = 2;
+
+    enum RequestType {
```
ulfjack: Why do you structure it like this? It seems like it would be easier to have separate methods for each rather than an enum and a single method. That also improves type safety for the response. The current design doesn't prevent returning dirents for a DIGEST request or vice versa.

designs/2019-02-26_RemoteFileSystem-For-Bazel.md
```diff
@@ -0,0 +1,225 @@
...
+        DirectoryEntries entries = 2;
+        FileStat filestat = 3;
+        OutputFile filecontent = 4;
+    }
+}
+
+//collection containing the names of all entities within the directory denoted by the path
+message DirectoryEntries {
+    //the relative path of the directory whose entries are being requested
+    string path = 1;
```
ulfjack: The client already knows the path, doesn't it?

designs/2019-02-26_RemoteFileSystem-For-Bazel.md
```diff
@@ -0,0 +1,225 @@
...
+
+    //The size of the file/directory
+    int64 size_bytes = 3;
+
+    //The last updated time of the file requested. analogous to utime on unix
+    //like file sysytems
+    int64 last_change_time = 4;
+
+    //Is the requested path a regular file ?
+    bool isFile = 5;
```
ulfjack: This should probably be an enum - even if you only care about files and directories _right now_, this'll eventually be expected to cover symlinks and junctions (windows directory symlinks) as well.

designs/2019-02-26_RemoteFileSystem-For-Bazel.md
```diff
@@ -0,0 +1,225 @@
...
+    //
+    // This field may be used by the server to provide the content of a file
+    // inline in an
+    // [ActionResult][google.devtools.remoteexecution.v1test.ActionResult] and
+    // avoid requiring that the client make a separate call to
+    // [ContentAddressableStorage.GetBlob] to retrieve it.
+    //
+    // The client SHOULD NOT assume that it will get raw content with any request,
+    // and always be prepared to retrieve it via `digest`.
+    bytes content = 3;
```
ulfjack: Returning the entire file as a single message is not going to scale to larger files - we have users who have multi-GB-sized files that this protocol will need to handle, even if it's uncommon.

designs/2019-02-26_RemoteFileSystem-For-Bazel.md
```diff
@@ -0,0 +1,225 @@
...
+}
+
+```
+
+# Caveats
+1. This implementation currently, causes a server restart on change of the source tree ( branch in git terminology ).
```
ulfjack: This is a design doc that shouldn't be specific to an implementation. Move this to wherever the impl is documented?

designs/2019-02-26_RemoteFileSystem-For-Bazel.md
JeromeDuval-Nam: Part of the design is to hook the remote file system into Bazel. Should we make a different design proposal for that part specifically? If so, should we reference the other one here and this one in the other one?

designs/2019-02-26_RemoteFileSystem-For-Bazel.md
JeromeDuval-Nam: I think that this is a design of the filesystem (Java) part of the proposal: how can we pass a version identifier that doesn't start invalidate everything that has been done on previous invocations. If we make a different proposal for that part of the remote file system I think it makes sense to move it to that other proposal.

designs/2019-02-26_RemoteFileSystem-For-Bazel.md
rahul-nitkkr: agreed, will be using some form of chunking to handle large files

rahul-nitkkr:
@ulfjack , could you please take a look at the updated proposal ?

designs/2019-02-26_RemoteFileSystem-For-Bazel.md
```diff
@@ -0,0 +1,170 @@
...
+We propose to have RemoteFileSystem which can talk to a HTTP/GRPC file server to satisfy the above listed requirements. The FileSystem at minimum needs to extend AbstractFileSystemWithCustomStat to intercept all Filesystem requests, and then appropiately delegate based on a list of either custom prefixes/any other delegation implementation.
+
+The BazelFileSystemModule needs to be extended to support the new Filesystem, and a set of Options need to be provided to provide the endpoint [ hosts / port ] as well as authentication options.
+
+RemoteFileSystem options should at the minimum be the following :
+
+### Startup Options
+
+1. **--enable_remote_filesystem** : flag to enable the intercation with remote file system, off by default
+2. **--remote_filesystem_type** : one of HTTP/GRPC...and FUSE ?
```
ulfjack: FUSE isn't really a remote file system because it looks like a local file system.

designs/2019-02-26_RemoteFileSystem-For-Bazel.md
ulfjack: I think a text description of what you're planning is sufficient.

designs/2019-02-26_RemoteFileSystem-For-Bazel.md
ulfjack: I don't think the exact Java classes are important, and I'm not sure we need a design doc for that part at all.

designs/2019-02-26_RemoteFileSystem-For-Bazel.md
```diff
@@ -0,0 +1,170 @@
...
+4. **--remote_filesystem_prefixes** : enables the delegating file system to decide what paths should be delegated to the RemoteFileSystem
+5. **--remote_filesystem_version** : optional parameter to point to version of the underlying repository to be used , defaults to master
+
+
+Any server attempting to support a remote file system implementation , will need to support the following requests :
+
+Operation | Endpoint | Returns
+----------|----------|--------
+Read Directory Entries |/bazel/dirents/:version/* | A List of directory entries
+Git File Input Stream | /bazel/data/:version/* | A stream of file bytes
```
johnynek: Do you mean Get here or Git? I assume the former.  If the latter, can you clarify how git is related?

rahul-nitkkr: This was a typo, fixing this. _Although, for Git, we have explored the option of running a server on top of gitlab [ our enterprise git client ] which aims to serve the api's mentioned below._

designs/2019-02-26_RemoteFileSystem-For-Bazel.md
```diff
@@ -0,0 +1,170 @@
...
+    //Type of the file requested corresponding to the path
+    enum FileType {
+        //a regular file
+        REGULAR_FILE = 0;
+
+        // a Direcotry
+        DIRECTORY = 1;
+
+        //Symlink or jucntion ( for windows )
+        SYMLINK = 2;
```
buchgr: Given that this is only for source files. Why support symlinks?

designs/2019-02-26_RemoteFileSystem-For-Bazel.md
```diff
@@ -0,0 +1,170 @@
...
+Operation | Endpoint | Returns
+----------|----------|--------
+Read Directory Entries |/bazel/dirents/:version/* | A List of directory entries
+Git File Input Stream | /bazel/data/:version/* | A stream of file bytes
+Get file Stat [ size, time_modified ]	| /bazel/filestat/:version/* | pair of size and time modified
+Get fast Digest	| /bazel/digest/:version/* | byte content of the file hash
+
+*Delegate provider implementation can be custom to all implementers , but at the very least they need to be a list of Path prefixes which can be resolved against a path to figure out the delegate.
+
+Alternative GRPC implementation proto is as follows. : remote_file_system.proto
```
buchgr: Why even have two protocols and not only support gRPC?

designs/2019-02-26_RemoteFileSystem-For-Bazel.md
```diff
@@ -0,0 +1,170 @@
...
+discussion thread: 
+---
+
+# Overview
+Bazel as a build tool natively expects files of the build action requested to be present on Disk. For a monorepo of a significant size, dumping out the entire repository for every single build , even with bazel's caching is going to be a significant overhead.
+
+Statistically, a repo which is ~2 GBs of source needs around 1-2 mins to dump out to disk. This proposal intends to address this particular issue by extending bazel to support a Remote File system , which can talk to any repository which supports Content addressable storage of source files to be able to provide on demand , requested files/contents hashes/file stats via HTTP/GRPC endpoints. Alternatively , a FUSE store can be specified as the type and associated mount point can be passed along as a startup option.
+
+# Existing Approaches
+Currently , the bazel server does not support OOTB any way to get around having files on the local disk where bazel server is running. Even with the remote [ BuildFarm ] setup , the local files are uploaded to the remote server for being actioned on.
```
buchgr: Please note that this is not required currently. Before uploading any source files to a remote execution system Bazel queries the system whether it already has these source files.

JeromeDuval-Nam:
> How hard would it be to implement this in Bazel and a simple server that can serve a git repository?

We have implemented the original version of this internally using our proprietary data store. The client part of this proposal (modifications to Bazel code) could definitely be pushed back out at as PR on a fork of Bazel.

The implementation of the backend part would be hard to spin out as is, since it requires a lot of our internal stuff. I think writing it backed by a Git repo and a naïve/simple implementation would probably be doable. I'm not sure of how quickly our internal open source procedure would approve the release of that code, since the process differs between open sourcing a new project and contributing to an existing project.

Would providing the Bazel part of the changes on their own help? The APIs are relatively descriptive of what they are supposed to achieve and the additional context of the Bazel code might be enough to illustrate how the design would work in practice. To set expectations for how long the creation of a new open source project might take, it will take a couple of weeks, maybe months depending on how thorough the vetting winds up being.

JeromeDuval-Nam:
> Thanks for the proposal!
> 
> I find it hard to review this proposal as is, as I don't have a good feel for the necessary APIs. Generally, I am all for having support for this natively in Bazel but before deciding on an API I think it would be hugely valuable to first have a prototype implementation for both a client and server that one can play around with.
> 
> How hard would it be to implement this in Bazel and a simple server that can serve a git repository?

Would it make sense to put the simple implementation of the server side of things in the example repository?

designs/2019-02-26_RemoteFileSystem-For-Bazel.md
```diff
@@ -0,0 +1,170 @@
...
+status: Draft
+reviewers: 
+  - ulfjack
+title: Remote File System for Bazel
+authors:
+  - royrah
+discussion thread: 
+---
+
+# Overview
```
EricBurnett: Meta-comment: I'd love to see this cross-reviewed with the [Remote Repository Cache](https://docs.google.com/document/d/1aagKnM8kRmMvLqSXasShsI-NYOBZYtgNrbYWLPwBoBg/edit) proposal, ( @jmillikin-stripe FYI). I think these proposals are solving sufficiently related problems that it'd be nice to see if one generalized remote API could suffice for both.

While I'm here, I'll also link my ["storage backend" vision](https://docs.google.com/document/d/17WJ4cz150IHeTgvJGxcbSGK1Zg2vsLpcgUSVffrwGfk/edit) doc, which is how I think about remote APIs clients like bazel may talk to. I think the contents are relevant here, but I'll let you read it and decide for yourself if you agree :).

designs/2019-02-26_RemoteFileSystem-For-Bazel.md
```diff
@@ -0,0 +1,170 @@
...
+The BazelFileSystemModule needs to be extended to support the new Filesystem, and a set of Options need to be provided to provide the endpoint [ hosts / port ] as well as authentication options.
+
+RemoteFileSystem options should at the minimum be the following :
+
+### Startup Options
+
+1. **--enable_remote_filesystem** : flag to enable the intercation with remote file system, off by default
+2. **--remote_filesystem_type** : one of HTTP/GRPC...and FUSE ?
+3. **--remote_filesystem_host** : host:port for the backend server for the filesystem requests in case of HTTP/GRPC
+4. **--remote_filesystem_prefixes** : enables the delegating file system to decide what paths should be delegated to the RemoteFileSystem
```
EricBurnett: Doing this via flag seems unscalable, for anything more than "exactly one path". If configuration of this type is needed, it seems it should go into its own configuration file bazel consumes.

Relates to the next flag as well - presumably if there are multiple paths, each also needs to be configured with a backing reference? (E.g. repository+hash). I see you have a "version" below, but I'm not sure how that'd work with multiple prefixes, and I'm not sure how the remote filesystem would then learn which repository each path refers to.

designs/2019-02-26_RemoteFileSystem-For-Bazel.md
```diff
@@ -0,0 +1,170 @@
...
+
+import "google/api/annotations.proto";
+import "build/bazel/remote/execution/v2/remote_execution.proto";
+
+//RPC endpoints for interfacing with a remotefile system server which is responsible
+//for resolving all file system interaction requested by bazel when running with remote
+//file system enabled.
+service RemoteFileSystem {
+
+    rpc FetchDigest(FileSystemRequest) returns (build.bazel.remote.execution.v2.Digest) {
```
EricBurnett: I'd suggest a few changes to the API here, to better scale to large repositories.

- Listing a directory, learning the digests, and learning the other stat details should be combined. You can see [Directory](https://github.com/bazelbuild/remote-apis/blob/master/build/bazel/remote/execution/v2/remote_execution.proto#L618) in the remote execution APIs for inspiration.
   - This is largely to minimize RPCs - a repo with 100k files would require >200k RPCs in the current framing (FetchDigest+FetchStat for each), whereas listing whole directories would likely be an order of magnitude less.
   - I'd also suggest including more metadata - which ones are directories, which are files, which are symlinks, etc.
- I'd suggest using the pattern of "learn digest of file, download by digest", rather than fetching content via a FileSystemRequest. That'd allow for sharing protos with the remote execution APIs, and possibly backends themselves. ByteStream+[BatchReadBlobs](https://github.com/bazelbuild/remote-apis/blob/master/build/bazel/remote/execution/v2/remote_execution.proto#L302), in this case. BatchReadBlobs is particularly relevant, since repos are mostly made up of a lot of small files, and it is _way_ faster to download those in batch RPCs than one by one, if downloading many files is required.

designs/2019-02-26_RemoteFileSystem-For-Bazel.md
```diff
@@ -0,0 +1,170 @@
...
+
+    rpc FetchContent(FileSystemRequest) returns (stream DataChunk) {
+        option (google.api.http) = { get: "/remotefs/file_content/{repo_version}/{path}" };
+    }
+}
+
+//a request to the remote file system is always comprised of 2 components
+//a path to resolve , and the version of the repository against which to find the path
+message FileSystemRequest {
+    //the relative path for this request, note that this relative to the WORKSPACE
```
EricBurnett: Should this not be relative to the base directory instead? Making it relative to the WORKSPACE requires the remote server to understand how bazel is laying out its own filesystem, and doesn't generalize to bazel mounting an arbitrary remote filesystem under an arbitrary path. That then seems too single-purpose for a remote API.

designs/2019-02-26_RemoteFileSystem-For-Bazel.md
```diff
@@ -0,0 +1,170 @@
...
+    string path = 1;
+
+    //version/commit of the repository against which the above path needs to be resolve.
```
EricBurnett: Similar to above, I think this needs to reference the repository itself. For the case where you have a monorepo and are only actually mounting that, you can pick any name you want for the repository if that suits your purposes, but for other usage of this API I'd assume it actually needs to reference the correct git repo (or other repo) that the version relates to.

designs/2019-02-26_RemoteFileSystem-For-Bazel.md
```diff
@@ -0,0 +1,170 @@
...
+    //Remote filesystems will be usually backed by a repository like Git, which rely on a version
+    //to fetch a file/content from.
+    //default value for this be master , and can be overridden with an empty string to if the underlying
```
EricBurnett: I'd recommend strongly *against* using master, or any sort of floating reference here - that'll lead to correctness issues if the the ref moves between two different RPCs.

jmillikin:
What's the current state of this proposal? It'd be nice to have a remote source tree, and having it built  into Bazel with gRPC is cleaner than implementing a custom FUSE filesystem.

designs/2019-02-26_RemoteFileSystem-For-Bazel.md
jmillikin: I think this is completely separate from the proposed remote repository cache. A remote source filesystem would allow Bazel to run without having a full copy of the source tree on local disk. A remote repository cache allows external dependencies to be fetched without hitting the original upstream.

designs/2019-02-26_RemoteFileSystem-For-Bazel.md
```diff
@@ -0,0 +1,170 @@
...
+# Suggested Implementation
+We propose to have RemoteFileSystem which can talk to a HTTP/GRPC file server to satisfy the above listed requirements. The FileSystem at minimum needs to extend AbstractFileSystemWithCustomStat to intercept all Filesystem requests, and then appropiately delegate based on a list of either custom prefixes/any other delegation implementation.
+
+The BazelFileSystemModule needs to be extended to support the new Filesystem, and a set of Options need to be provided to provide the endpoint [ hosts / port ] as well as authentication options.
+
+RemoteFileSystem options should at the minimum be the following :
+
+### Startup Options
+
+1. **--enable_remote_filesystem** : flag to enable the intercation with remote file system, off by default
```
jmillikin: Using the existing remote cache flags as an example, you could have the presence of `--remote_filesystem=grpc://some-address` enable the remote FS feature.

designs/2019-02-26_RemoteFileSystem-For-Bazel.md
```diff
@@ -0,0 +1,170 @@
...
+2. **--remote_filesystem_type** : one of HTTP/GRPC...and FUSE ?
+3. **--remote_filesystem_host** : host:port for the backend server for the filesystem requests in case of HTTP/GRPC
+4. **--remote_filesystem_prefixes** : enables the delegating file system to decide what paths should be delegated to the RemoteFileSystem
+5. **--remote_filesystem_version** : optional parameter to point to version of the underlying repository to be used , defaults to master
```
jmillikin: I think you'll want to leave this up to the backend. There are many ways to encode repository versions, and not all of them are amenable to being a single static flag.

For example, a backend based on Git might use `--remote_filesystem_prefix=/acmecorp/src/`, and then the backend would resolve `/acmecorp/src/by-ref/da39a3ee5e6b4b0d3255bfef95601890afd80709/` to the source tree at that revision.

designs/2019-02-26_RemoteFileSystem-For-Bazel.md
jmillikin: You could have something like `--remote_filesystem_prefix=/acmecorp/src/ --remote_filesystem_prefix=/acmecorp/obj/`, and have the full path be sent to the backend. This would require the backend to be prefix-dependent, or at least prefix-_aware_.

designs/2019-02-26_RemoteFileSystem-For-Bazel.md
```diff
@@ -0,0 +1,170 @@
...
+option java_outer_classname = "RemoteFileSystemProtos";
+option java_multiple_files = true;
+
+import "google/api/annotations.proto";
+import "build/bazel/remote/execution/v2/remote_execution.proto";
+
+//RPC endpoints for interfacing with a remotefile system server which is responsible
+//for resolving all file system interaction requested by bazel when running with remote
+//file system enabled.
+service RemoteFileSystem {
```
jmillikin: I recommend using FUSE as a starting point for this API, as it's been pretty well optimized for common filesystem patterns. You'll want to avoid round-tripping file paths as Bazel walks the tree, and will want to use streams for anything that has an unbounded size (like directory listing).

Rough sketch:

```
service Filesystem {
  rpc LookupTree(LookupRequest) returns (LookupResponse);
}

message LookupRequest {
  string instance_name;
  string path; // directory path to look up the digest for
}
message LookupResponse {
  Digest root_digest; // can be passed to `ContentAddressableStorage.GetTree`
}
```

designs/2019-02-26_RemoteFileSystem-For-Bazel.md
jmillikin: Is it important to support arbitrary prefixes? Presumably this would be used mostly for internal CI systems, or at least for well-controlled environments where both Bazel and the remote filesystem are run by the same people.

designs/2019-02-26_RemoteFileSystem-For-Bazel.md
```diff
@@ -0,0 +1,170 @@
...
+
+    //the list of entries in this particulr directory path
+    repeated string entries = 2;
+
+}
+
+//a 'FileStat' represents the status entries for a file : File status: mode, mtime, size, etc.
+//the remote implementer of the FileSystem service needs to provide the fields to accurately define
+//the state of a file in the FileSystem
+message FileStat {
```
jmillikin: I think most of this `FileStat` message is unnecessary -- the content of `build.bazel.remote.execution.v2.FileNode` is sufficient to identify files to Bazel, and things like modification times aren't used by Bazel to decide which actions should be re-executed.

jin:
Closing this as the discussion has gone stale for a while now.

PR #103, opened 2019-04-05
Added draft proposal - Propagate tags from the targets to the actions

ishikhman:
part of https://github.com/bazelbuild/bazel/issues/7766


ishikhman:
> Thanks
> 
> Next step is to send it to bazel-dev@

Thanks! Could you please merge this PR? I do not have write permissions.

PR #105, opened 2019-04-12
Change output path proposal to Under Review.

gregestren:
According to [design policy](https://bazel.build/designs/index.html), this PR is just to switch the status for "In Review", not to actually approve the review.

After this PR is in I'll start a bazel-dev discussion to kick off review proper.


gregestren:
@lberki and @serynth  - which email addresses / GitHub handles / whatever identifiers would you like me to use for you on the doc?

gregestren:
Ping @lberki - does this look ready to start review?

aiuto:
Since this PR has the link to  the document, it is already something that people can see.
I don't see any point in holding up this PR, since it simply starts the discussion. It does not have to mean that it also starts a countdown clock for @lberki. I would go ahead and merge.

PR #116, opened 2019-05-06
updated a url to a local repo for design document

genrym:
The link to the design document was to the forked repository.
Updated it to be as relative to the repo.

@aehlig Should I open a discussion in `bazel-dev` about the design?


laurentlb:
The workflow is described here: https://bazel.build/designs/#workflow

Yes, please start a discussion on bazel-dev.
Thank you.

PR #124, opened 2019-06-07
Mark authentication design as approved

aehlig:
...as was decided in the Starlark meeting at June 6, 2019.
Also add a clarifying statement that http_archive will by
default honor a `.netrc` file even though this will not
be part of the build API, but instead an implementation
detail of the corresponding Starlark rule.


aehlig:
Superseded by https://github.com/bazelbuild/proposals/pull/129

PR #144, opened 2019-10-24
tool visibility: add an example

aehlig:
In the background section of the tool-visibility proposal,
add a simplified example showing how visibility rules make it
hard to evolve the interface of an internal tool. A real-world
example was already linked in the background section but is
more involved due to additional indirections.

### DIFF
--- a/designs/2019-10-15-tool-visibility.md
+++ b/designs/2019-10-15-tool-visibility.md
@@ -106,6 +106,9 @@ rule definition). In this way, it will be possible to restrict the
 visibility of a tool to the package containing the rule definition,
 while still allowing the rule to be used everywhere.
 
+In the previous example, this means the visibility could be restricted
+to `//foo/rule:__pkg__` where the rule is defined, instead of every user
+of the rule like `//bar:myapp`.
 
 # Backward-compatibility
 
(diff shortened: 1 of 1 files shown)


designs/2019-10-15-tool-visibility.md
```diff
@@ -55,6 +55,46 @@ In fact, that tool motivated the [first
...
+```
+
+Then `//bar:myapp` implicitly depends on `//foo/tool:foobuilder`, so
+`//foo/tool:foobuilder` must be visible by `//bar:myapp`. This visibility
+allows targets in `//bar` to also use `foobuilder` directly, even if it
+is intended to be an implementation detail of `foo_binary`. This unrestricted
+use makes it hard for the `//foo` package to evolve the command-line interface
+of `foobuilder`.
+
+
```
laurentlb: In the proposal section, maybe add a new paragraph:

> In the previous example, this means that the visibility of `//foo/tool:foobuilder` could be restricted to `//foo/tool` (where the rule is defined), instead of being visible to every user of the rule.

aehlig: done.

PR #149, opened 2019-11-11
Add draft proposal to expose target platform constraints to Starlark

katre:
This is a draft: please comment only on formatting, grammar, etc, and not on the content. Once accepted I'll mark it as Under Review and begin to publicize.

### DIFF
--- a/designs/2019-11-11-target-platform-constraints.md
+++ b/designs/2019-11-11-target-platform-constraints.md
@@ -1,0 +1,137 @@
+---
+created: 2019-11-11
+last updated: 2019-11-11
+status: Draft
+reviewers:
+  - aragos
+  - laurentlb
+title: Exposing Target Platform Constraints
+authors:
+  - katre
+---
+
+
+# Abstract
+
+Rule authors occasionally need information about the target platform in a rule
+implementation. This can be achieved currently using either custom toolchains or
+by using implicit attributes with a `select()` call. This proposal aims to allow
+rule authors the information they need in a simpler manner.
+
+# Background
+
+Every configured target has a target platform that the outputs are compatible
+with. This target platform may be set (either by default or by a command-line
+flag) when the build begins, or may be changed as a result of a configuration
+transition.
+
+A current workaround to determine the target platform is to check
+[`ctx.configuration.host_path_separator`](https://github.com/bazelbuild/bazel/issues/9209),
+or to add [an implicit attribute that selects on a relevant platform
+constraint](https://github.com/bazelbuild/bazel/issues/9209#issuecomment-539239447).
+However, both of these are inexact and add additional overhead for rule authors.
+
+# Current Status
+
+The target platform information, as a [PlatformInfo
+provider](https://source.bazel.build/bazel/+/master:src/main/java/com/google/devtools/build/lib/analysis/platform/PlatformInfo.java),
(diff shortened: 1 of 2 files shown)


designs/2019-11-11-target-platform-constraints.md
```diff
@@ -0,0 +1,137 @@
...
+```
+
+# Proposal: A check method that takes a constraint value label
+
+One option is to expose a method (again, on either `SkylarkRuleContextApi` or
+`ToolchainContextApi`) to check whether a given constraint value is present.
+Showing only the rule context version:
+
+```py
+if ctx.target_platform_has_constraint('@bazel_platforms//os:windows'):
```
aragos: Do we expect additional `target_platform` or `execution_platform` related accessors/getters at some point? If yes it may be better to structure this as `ctx.target_platform.has_constraint`.

designs/2019-11-11-target-platform-constraints.md
```diff
@@ -0,0 +1,137 @@
...
+create a similar method in
+[`ConstraintCollection`](https://source.bazel.build/bazel/+/master:src/main/java/com/google/devtools/build/lib/analysis/platform/ConstraintCollection.java?q=ConstraintCollection)
+to perform the actual check.
+
+## Alternative 1: A list of all constraint values
+
+The simplest option to enable this feature would be to simply to expose all
+[`ConstraintValueInfoApi`](https://source.bazel.build/bazel/+/master:src/main/java/com/google/devtools/build/lib/skylarkbuildapi/platform/ConstraintValueInfoApi.java)
+values from the current target platform's `PlatformInfo`. This would be fairly
+easy to add as a new attribute on either `SkylarkRuleContextApi' or
```
aragos: Replace ' with `? :)

designs/2019-11-11-target-platform-constraints.md
```diff
@@ -0,0 +1,137 @@
...
+`ToolchainContextApi`, depending on which form of rule code is preferred:
+
+```py
+# Via the rule context directly:
+toolchain = ctx.toolchains['//tools/cpp:toolchain_type']
```
aragos: This line seems superfluous?

designs/2019-11-11-target-platform-constraints.md
```diff
@@ -0,0 +1,137 @@
...
+  attrs = {
+    ...
+    '_windows_constraint': attr.label(default = '@bazel_platforms//os:windows'),
+  },
+)
+```
+
+This would ensure that rules can only access constraints that they are allowed
+to establish dependencies on, at the cost of requiring more complexity for rule
+authors.
```
aragos: The more I think about this the more I like it. :)

designs/2019-11-11-target-platform-constraints.md
katre: Let's discuss that on the review thread.

designs/2019-11-11-target-platform-constraints.md
katre: Done.

designs/2019-11-11-target-platform-constraints.md
katre: I added it to demonstrate the issue with the second version, getting both toolchain providers and constraint providers from the `ctx.toolchains` object.

PR #150, opened 2019-11-13
Mark file visibility as implemented

aehlig:
It has been approved and the flag was added in
https://github.com/bazelbuild/bazel/commit/2bb1bf9422ba22d0e1ab15363664339f022b29a1


aehlig:
> If not done, please create an issue to track the incompatible flag

The flag is tracked in (the already created) issue https://github.com/bazelbuild/bazel/issues/10225

PR #160, opened 2020-01-14
Unify remote downloader proposals (remote repo cache, remote fetch api)

jmillikin-stripe:
Per email discussion, this takes the two existing proposals for remote downloading of external dependencies ([Remote Repository Cache](https://docs.google.com/document/d/1aagKnM8kRmMvLqSXasShsI-NYOBZYtgNrbYWLPwBoBg/view) and [Remote 'Fetch' API](https://docs.google.com/document/d/10ari9WtTTSv9bqB_UU-oe2gBtaAA7HyQgkpP-RFP80c/view)) and merges them into the Bazel proposals repository.

The Protobuf portions of the merged proposal are copied from the Fetch API doc, because that's the version that got approved by Bazel + RBE devs. The Bazel-specific portions (e.g. implementation plan) plus a bunch of the background/motivation is from the Remote Repository Cache doc.

I tried my best to keep both the spirit and details of both documents intact. If there's anywhere the authors of the Fetch API doc would prefer to use the original wording, please comment and I'll update this PR to do so.

Once this is merged, I'll send a PR to the remote-apis repo to make the new API available.

R: @buchgr @dslomov
CC: @EricBurnett @sstriker

### DIFF
--- a/designs/2020-01-14-remote-downloads.md
+++ b/designs/2020-01-14-remote-downloads.md
@@ -119,8 +119,8 @@ reasonable period of time thereafter.
 
 In the event that a client receives a reference to content that is no longer
 present in the CAS, it **MAY** re-issue the request with
-`oldest_content_accepted` set to a value later than the original attempt, to
-induce a re-fetch from origin.
+`oldest_content_accepted` set to a timestamp more recent than the original
+attempt, to induce a re-fetch from origin.
 
 Servers **MAY** cache fetched content and reuse it for subsequent requests,
 subject to `oldest_content_accepted`. Servers **MAY** fetch content that they
@@ -503,7 +503,7 @@ message Qualifier {
 
 1. Refactor the existing `HttpDownloader` path to support non-HTTP
    implementations. While the default implementation will match existing
-   behavior, it will be possible modules to inject a new `Downloader` with
+   behavior, it will be possible for modules to inject a new `Downloader` with
    custom logic.
 
 2. Add a new `--experimental_remote_downloader=ADDRESS` flag to `RemoteModule`,
@@ -511,8 +511,8 @@ message Qualifier {
    implementation should leverage existing support code for mTLS credentials
    and UNIX socket proxying.
 
-3. After the implementation has received enough use to consider stable, rename
-   the flag to `--remote_downloader`.
+3. After the implementation has received enough use to be considered stable,
+   rename the flag to `--remote_downloader`.
 
 ## Backward-compatibility
 
@@ -543,7 +543,7 @@ straightforward to implement a caching intermediate proxy for downloads with the
 
 ## Direct downloads
 
-The first version of the Remote Reposiotry Cache proposal had the server stream
+The first version of the Remote Repository Cache proposal had the server stream
 downloads directly, as in `ByteStream`:
 
 ```proto
--- a/designs/2020-01-14-remote-downloads.md
+++ b/designs/2020-01-14-remote-downloads.md
@@ -113,14 +113,13 @@ service Fetcher {
 }
 ```
 
-If a request is successful then servers **SHOULD** ensure the referenced assets
+If a request is successful then servers **MUST** ensure the referenced assets
 are present in the CAS at the time of response and (if supported) for a
 reasonable period of time thereafter.
 
-In the event that a client receives a reference to content that is no longer
-present in the CAS, it **MAY** re-issue the request with
-`oldest_content_accepted` set to a timestamp more recent than the original
-attempt, to induce a re-fetch from origin.
+The client **MAY** request a resource be re-fetched from its origin by
+setting the request field `oldest_content_accepted` to a timestamp more recent
+than the previous attempt.
 
 Servers **MAY** cache fetched content and reuse it for subsequent requests,
 subject to `oldest_content_accepted`. Servers **MAY** fetch content that they
@@ -163,9 +162,16 @@ message FetchBlobRequest {
   string instance_name = 1;
 
   // The timeout for the underlying fetch, if content needs to be retrieved from
-  // origin. This value is allowed to exceed the RPC deadline, in which case the
-  // server *SHOULD* keep the fetch going after the RPC completes, to be made
-  // available for future Fetch calls.
+  // origin.
+  //
+  // If unset, the server *MAY* apply an implementation-defined timeout.
+  //
+  // If set, and the user-provided timeout exceeds the RPC deadline, the server
+  // *SHOULD* keep the fetch going after the RPC completes, to be made
+  // available for future Fetch calls. The server may also enforce (via clamping
+  // and/or an INVALID_ARGUMENT error) implementation-defined minimum and
+  // maximum timeout values.
+  //
   // If this timeout is exceeded on an attempt to retrieve content from origin
   // the client will receive DEADLINE_EXCEEDED in [FetchBlobResponse.status].
   google.protobuf.Duration timeout = 2;
@@ -175,6 +181,8 @@ message FetchBlobRequest {
   // Upon retries of Fetch requests that cannot be completed within a single RPC,
   // clients *SHOULD* provide the same value for subsequent requests as the
   // original, to simplify combining the request with the previous attempt.
+  //
+  // If unset, the client *SHOULD* accept content of any age.
   google.protobuf.Timestamp oldest_content_accepted = 3;
 
   // The URI(s) of the content to fetch. These may be resources that the server can
@@ -411,12 +419,14 @@ message PushDirectoryResponse { /* empty */ }
 
 ## Qualifiers
 
-Qualifiers are used to disambiguate or sub-select content that shares a URI.
-This may include specifying a particular commit or branch, in the case of URIs
-referencing a repository; they could also be used to specify a particular
-subdirectory of a repository or tarball. Qualifiers may also be used to ensure
-content matches what the client expects, even when there is no ambiguity to be
-had -- for example, a qualifier specifying a checksum value.
+Qualifiers are used to disambiguate, sub-select, or validate content that shares
+a URI. This may include specifying a particular commit or branch, in the case of
+URIs referencing a repository; they could also be used to specify a particular
+subdirectory of a repository or tarball.
+
+Qualifiers may also be used to ensure content matches what the client expects,
+even when there is no ambiguity to be had -- for example, a qualifier specifying
+a checksum value.
 
 For example, a particular subdirectory of a particular branch
 of a git repository might be identified as:
--- a/designs/2020-01-14-remote-downloads.md
+++ b/designs/2020-01-14-remote-downloads.md
@@ -25,6 +25,13 @@ and security characteristics in some cases. This proposal expands the remote
 execution APIs to support downloading inputs from external sources, allowing the
 build tool to offload expensive or privileged network requests.
 
+Note: The Protobuf components of this proposal have been submitted (with slight
+changes) in [bazelbuild/remote-apis#112][remote-apis-pr-112]
+as [`build/bazel/remote/asset/v1/remote_asset.proto`][remote_asset.proto].
+
+[remote-apis-pr-112]: https://github.com/bazelbuild/remote-apis/pull/112
+[remote_asset.proto]: https://github.com/bazelbuild/remote-apis/blob/master/build/bazel/remote/asset/v1/remote_asset.proto
+
 # Background
 
 When a build tool is calculating which actions to execute as part of a build, it
--- a/designs/2020-01-14-remote-downloads.md
+++ b/designs/2020-01-14-remote-downloads.md
@@ -1,0 +1,589 @@
+---
+created: 2019-04-18
+last updated: 2020-01-20
+status: Approved
+reviewers:
+  - aehlig
+  - buchgr
+title: Remote Downloads
+authors:
+  - EricBurnett
+  - jmillikin
+  - sstriker
+discussion threads:
+  - https://groups.google.com/forum/#!topic/remote-execution-apis/Gyzvz7WkQPE
+  - https://groups.google.com/forum/#!topic/remote-execution-apis/kTcjXrijWpU
+---
+
+# Abstract
+
+The [Remote Execution APIs](https://github.com/bazelbuild/remote-apis) are a
+set of common gRPC APIs covering remote execution and content-addressed storage,
+primarily intended for use by build tools. Currently these APIs require all
+execution inputs to be uploaded by the client, which has undesirable performance
+and security characteristics in some cases. This proposal expands the remote
+execution APIs to support downloading inputs from external sources, allowing the
+build tool to offload expensive or privileged network requests.
+
+# Background
+
+When a build tool is calculating which actions to execute as part of a build, it
+may need to download input files from external sources -- common external
+dependencies include compiler toolchains, third-party libraries, test data
+snapshots, and so on. There are three important reasons why the build tool
+should not be the thing doing the actual downloading:
+
+* The external dependencies might be hosted on a third-party service, and build
+  reliability will be degraded if that service has an outage or temporarily
+  throttles downloads. Downloading via an intermediate service allows the
+  injection of a caching proxy layer.
+
+* Some organizations would like to enforce policies on use of third-party code,
+  and a build tool that directly downloads external dependencies will bypass
+  technical controls that implement these policies.
+
+* The build tool might have a low-quality network connection, such that
+  downloading and re-uploading dependencies is a significant fraction of the
+  build time. If the remote execution service is hosted in a datacenter, then
+  it can fetch dependencies much faster than the client can.
+
+This proposal is derived from two previous documents:
+
+* [Remote Repository Cache](https://docs.google.com/document/d/1aagKnM8kRmMvLqSXasShsI-NYOBZYtgNrbYWLPwBoBg/view)
+  ([discussion](https://groups.google.com/forum/#!topic/remote-execution-apis/Gyzvz7WkQPE))
+  is a Bazel-specific gRPC API for downloading external dependencies. It lays
+  out the basic design goals and implementation strategy for a remote downloader.
+
+* [Remote 'Fetch' API](https://docs.google.com/document/d/10ari9WtTTSv9bqB_UU-oe2gBtaAA7HyQgkpP-RFP80c/view)
+  ([discussion](https://groups.google.com/forum/#!topic/remote-execution-apis/kTcjXrijWpU))
+  generalized the above API to be non-Bazel-specific so that it could be accpted
+  into the Remote Execution APIs, and expanded it to cover use cases of
+  the BuildStream project.
+
+# Protocol
+
+The new API is built on top of the existing `ContentAddressableStorage` service,
+but has its own gRPC service names so that traffic can be routed independently.
+
+Resources are identified by URI(s) and expected qualifiers. Qualifiers are
+arbitrary key-value pairs, with certain "well-known" keys documented in the
+[**Qualifiers**](#qualifiers)] section of this document.
+
+A minimal compliant Fetcher implementation might support only pushed content and
+return `NOT_FOUND` for any resource that was not pushed first. Alternatively,
+a compliant implementation may choose to not support Push and only return
+resources that can be retrieved from their origin.
+
+## Resource Identifiers
+
+Clients **MUST** supply one or more URIs in requests; servers **MAY** match any
+of the supplied URIs. All URIs in a request are expected to identify the same
+content, and thus share the same Qualifiers.
+
+Support for specific URI schemes is implementation-dependent. Servers may choose
+to support only URLs, only URNs, only other URIs for which the server and client
+agree upon semantics thereof, or any mixture of the above.
+
+Multiple URIs may be used to refer to the same content. For example, the same
+tarball may exist at multiple mirrors and thus be retrievable from multiple
+URLs. When URLs are used, these should refer to actual content as Fetch API
+service implementations may choose to fetch the content directly from the
+origin. For example, the HEAD of a git repository's active branch can be
+referred to as:
+
+```
+uri: "https://github.com/bazelbuild/remote-apis.git"
+```
+
+URNs may be used as opaque content identifiers, for instance by using the `uuid`
+namespace: `urn:uuid:f81d4fae-7dec-11d0-a765-00a0c91e6bf6`. This is most
+applicable to content that is injected into the CAS via the Push API, where the
+URN serves as an agreed-upon key, but carries no other inherent meaning.
+
+## The 'Fetcher' API
+
+The `Fetcher` gRPC service injects entries into the content-addressable storage
+by downloading data from external sources. If successful, the server returns the
+`Digest` of the fetched content so that the client can retrieve it from the CAS.
+
+```proto
+service Fetcher {
+  rpc FetchBlob(FetchBlobRequest) returns (FetchBlobResponse);
+  rpc FetchDirectory(FetchDirectoryRequest) returns (FetchDirectoryResponse);
+}
+```
+
+If a request is successful then servers **SHOULD** ensure the referenced assets
+are present in the CAS at the time of response and (if supported) for a
+reasonable period of time thereafter.
+
+In the event that a client receives a reference to content that is no longer
+present in the CAS, it **MAY** re-issue the request with
+`oldest_content_accepted` set to a value later than the original attempt, to
+induce a re-fetch from origin.
+
+Servers **MAY** cache fetched content and reuse it for subsequent requests,
+subject to `oldest_content_accepted`. Servers **MAY** fetch content that they
+do not already have cached, for any URIs they support.
+
+Servers **MUST** reject requests containing unknown Qualifiers. Servers **MAY**
+validate the fetched content against the requested qualifiers. Servers **MAY**
+have different validation policies for assets obtained from different sources,
+for example by special-casing assets injected into the CAS via the `Pusher` API.
+
+Servers **MAY** transform assets as part of the fetch. For example a tarball
+fetched by `FetchDirectory` might be unpacked, or a Git repository fetched by
+`FetchBlob` might be passed through `git-archive`.
+
+Errors encountered processing the request will be returned as gRPC errors with
+the following codes:
+
+* `INVALID_ARGUMENT`: One or more arguments to the RPC were invalid.
+
+* `RESOURCE_EXHAUSTED`: There is insufficient quota of some resource to
+  perform the requested operation. The client may retry after a delay.
+
+* `UNAVAILABLE`: Due to a transient condition the operation could not be
+  completed. The client should retry.
+
+* `INTERNAL`: An internal error occurred while performing the operation. The
+  client should retry.
+
+* `DEADLINE_EXCEEDED`: The fetch could not be completed within the given
+  RPC deadline. The client should retry for at least as long as the value
+  provided in `request.timeout`.
+
+Errors encountered while fetching assets will be returned in the responses'
+`status` field. See the relevant response types for details.
+
+### FetchBlobRequest
+
+```proto
+message FetchBlobRequest {
+  string instance_name = 1;
+
+  // The timeout for the underlying fetch, if content needs to be retrieved from
+  // origin. This value is allowed to exceed the RPC deadline, in which case the
+  // server *SHOULD* keep the fetch going after the RPC completes, to be made
+  // available for future Fetch calls.
+  // If this timeout is exceeded on an attempt to retrieve content from origin
+  // the client will receive DEADLINE_EXCEEDED in [FetchBlobResponse.status].
+  google.protobuf.Duration timeout = 2;
+
+  // The oldest content the client is willing to accept, as measured from the
+  // time it was Push'd or when the underlying retrieval from origin was started.
+  // Upon retries of Fetch requests that cannot be completed within a single RPC,
+  // clients *SHOULD* provide the same value for subsequent requests as the
+  // original, to simplify combining the request with the previous attempt.
+  google.protobuf.Timestamp oldest_content_accepted = 3;
+
+  // The URI(s) of the content to fetch. These may be resources that the server can
+  // directly fetch from origin, in which case multiple URIs *SHOULD* represent
+  // the same content available at different locations (such as an origin and
+  // secondary mirrors). These may also be URIs for content known to the server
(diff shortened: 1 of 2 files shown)


buchgr:
If nobody has any comments I'll merge this in two business days.

designs/2020-01-14-remote-downloads.md
```diff
@@ -0,0 +1,589 @@
...
+  upstreams. It would be difficult to make Bazel send integrity headers in some
+  proxy environments, and not in others.
+
+* This proposal does not prevent a simplified HTTP-based cache from being
+  implemented at a later date. If demand for that capability exists, it should
+  be proposed separately.
+
+## Direct downloads
+
+The first version of the Remote Reposiotry Cache proposal had the server stream
```
buchgr: s/Reposiotry/Repository/

designs/2020-01-14-remote-downloads.md
```diff
@@ -0,0 +1,589 @@
...
+execution APIs to support downloading inputs from external sources, allowing the
+build tool to offload expensive or privileged network requests.
+
+# Background
+
+When a build tool is calculating which actions to execute as part of a build, it
+may need to download input files from external sources -- common external
+dependencies include compiler toolchains, third-party libraries, test data
+snapshots, and so on. There are three important reasons why the build tool
+should not be the thing doing the actual downloading:
```
ulfjack: Another is to avoid DOSing the server providing these if there are a large number of concurrent build tool invocations for whatever reason.

designs/2020-01-14-remote-downloads.md
```diff
@@ -0,0 +1,589 @@
...
+
+* The external dependencies might be hosted on a third-party service, and build
+  reliability will be degraded if that service has an outage or temporarily
+  throttles downloads. Downloading via an intermediate service allows the
+  injection of a caching proxy layer.
+
+* Some organizations would like to enforce policies on use of third-party code,
```
ulfjack: Technically, the build tool could implement the policies, or you could setup a firewall configuration to enforce the policies.

designs/2020-01-14-remote-downloads.md
```diff
@@ -0,0 +1,589 @@
...
+}
+```
+
+If a request is successful then servers **SHOULD** ensure the referenced assets
+are present in the CAS at the time of response and (if supported) for a
+reasonable period of time thereafter.
+
+In the event that a client receives a reference to content that is no longer
+present in the CAS, it **MAY** re-issue the request with
+`oldest_content_accepted` set to a value later than the original attempt, to
```
ulfjack: s/set to a value later/set to a timestamp more recent/?

designs/2020-01-14-remote-downloads.md
```diff
@@ -0,0 +1,589 @@
...
+message FetchBlobRequest {
+  string instance_name = 1;
+
+  // The timeout for the underlying fetch, if content needs to be retrieved from
+  // origin. This value is allowed to exceed the RPC deadline, in which case the
+  // server *SHOULD* keep the fetch going after the RPC completes, to be made
+  // available for future Fetch calls.
+  // If this timeout is exceeded on an attempt to retrieve content from origin
+  // the client will receive DEADLINE_EXCEEDED in [FetchBlobResponse.status].
+  google.protobuf.Duration timeout = 2;
```
ulfjack: What should happen if this is not set?

designs/2020-01-14-remote-downloads.md
```diff
@@ -0,0 +1,589 @@
...
+
+  // The oldest content the client is willing to accept, as measured from the
+  // time it was Push'd or when the underlying retrieval from origin was started.
+  // Upon retries of Fetch requests that cannot be completed within a single RPC,
+  // clients *SHOULD* provide the same value for subsequent requests as the
+  // original, to simplify combining the request with the previous attempt.
+  google.protobuf.Timestamp oldest_content_accepted = 3;
```
ulfjack: Can this be unset or set to 0? What should happen in that case?

designs/2020-01-14-remote-downloads.md
```diff
@@ -0,0 +1,589 @@
...
+
+  // The URI(s) of the content to fetch. These may be resources that the server can
+  // directly fetch from origin, in which case multiple URIs *SHOULD* represent
+  // the same content available at different locations (such as an origin and
+  // secondary mirrors). These may also be URIs for content known to the server
+  // through other mechanisms, e.g. pushed via the [Pusher] API.
```
ulfjack: Why would you call 'fetch' on something you previously pushed?

designs/2020-01-14-remote-downloads.md
```diff
@@ -0,0 +1,589 @@
...
+  //   requested an asset from a disallowed origin.
+  // * `ABORTED`: The operation could not be completed, typically due to a failed
+       consistency check.
+  google.rpc.Status status = 1;
+
+  // The uri from the request that resulted in a successful retrieval, or from
+  // which the error indicated in `status` was obtained.
+  string uri = 2;
+
+  // Any qualifiers known to the server and of interest to clients.
```
ulfjack: Can you elaborate or give an example?

designs/2020-01-14-remote-downloads.md
```diff
@@ -0,0 +1,589 @@
...
+  // The result of the fetch, if the status had code `OK`.
+  // The digest of the file's contents, available for download through the CAS.
+  Digest blob_digest = 5;
+}
+```
+
+### FetchDirectoryRequest
+
+```proto
+message FetchDirectoryRequest {
```
ulfjack: How's this different from FetchBlobRequest?

designs/2020-01-14-remote-downloads.md
```diff
@@ -0,0 +1,589 @@
...
+  // Any qualifiers known to the server and of interest to clients.
+  repeated Qualifier qualifiers = 3;
+
+  // A minimum timestamp the content is expected to be available through.
+  // Servers *MAY* omit this field, if not known with confidence.
+  google.protobuf.Timestamp expires_at = 4;
+
+  // The result of the fetch, if the status had code `OK`.
+  // the root digest of a directory tree, suitable for fetching via
+  // [ContentAddressableStorage.GetTree].
```
ulfjack: Why would this be required? IIUC, the CAS generally doesn't know whether something may or may not be a directory, and this now seems to indicate that the fetch server should know?

designs/2020-01-14-remote-downloads.md
```diff
@@ -0,0 +1,589 @@
...
+
+  // Qualifiers sub-specifying the content that is being pushed - see comments on
+  // [Qualifier]. The same qualifiers apply to all URIs.
+  repeated Qualifier qualifiers = 3;
+
+  // A time after which this content should stop being returned via Fetch.
+  // Servers *MAY* expire content early, e.g. due to storage pressure.
+  google.protobuf.Timestamp expire_at = 4;
+
+  // The blob to associate.
```
ulfjack: Where is the blob? Is it supposed to be stored in the CAS first?

designs/2020-01-14-remote-downloads.md
```diff
@@ -0,0 +1,589 @@
...
+
+### PushDirectoryResponse
+
+```proto
+message PushDirectoryResponse { /* empty */ }
+```
+
+## Qualifiers
+
+Qualifiers are used to disambiguate or sub-select content that shares a URI.
```
ulfjack: I think the comment below about ambiguity refers to the use of "disambiguate" here. How about instead writing that the qualifiers are used to "disambiguate, sub-select, or validate content"?

designs/2020-01-14-remote-downloads.md
```diff
@@ -0,0 +1,589 @@
...
+* Qualifier keys are expected to be unique. Semantics of repeated keys are
+  unspecified at this time.
+* Qualifiers may be present in any order.
+* No attempt is made to distinguish between "Standard" and "Nonstandard"
+  qualifiers. In this we follow the guidance of
+  https://tools.ietf.org/html/rfc6648. It is anticipated that a standard set of
+  keys and associated value semantics will be developed as this proposal lands,
+  and added to the documentation as appropriate.
+* In cases where the semantics of the request are not immediately clear from the
+  URL and/or qualifiers - e.g. dictated by URL scheme - it is recommended to use
```
ulfjack: Can you give an example?

designs/2020-01-14-remote-downloads.md
```diff
@@ -0,0 +1,589 @@
...
+message PushBlobRequest {
+  string instance_name = 1;
+
+  // The URI(s) of the content to associate. If multiple URIs are specified, the
+  // pushed content will be available to fetch by specifying any of them.
+  repeated string uris = 2;
+
+  // Qualifiers sub-specifying the content that is being pushed - see comments on
+  // [Qualifier]. The same qualifiers apply to all URIs.
+  repeated Qualifier qualifiers = 3;
```
ulfjack: Did you consider representing expiration and digests as qualifiers?

designs/2020-01-14-remote-downloads.md
```diff
@@ -0,0 +1,589 @@
...
+}
+```
+
+# Bazel
+
+## Implementation plan
+
+1. Refactor the existing `HttpDownloader` path to support non-HTTP
+   implementations. While the default implementation will match existing
+   behavior, it will be possible modules to inject a new `Downloader` with
```
ulfjack: s/it will be possible modules/it will be possible for modules/

designs/2020-01-14-remote-downloads.md
```diff
@@ -0,0 +1,589 @@
...
+   custom logic.
+
+2. Add a new `--experimental_remote_downloader=ADDRESS` flag to `RemoteModule`,
+   which injects a `GrpcRemoteDownloader` backed by a remote endpoint. The
+   implementation should leverage existing support code for mTLS credentials
+   and UNIX socket proxying.
+
+3. After the implementation has received enough use to consider stable, rename
```
ulfjack: to *be* considered stable

designs/2020-01-14-remote-downloads.md
```diff
@@ -0,0 +1,589 @@
...
+```
+
+# Bazel
+
+## Implementation plan
+
+1. Refactor the existing `HttpDownloader` path to support non-HTTP
+   implementations. While the default implementation will match existing
+   behavior, it will be possible modules to inject a new `Downloader` with
+   custom logic.
```
ulfjack: This item doesn't seem to be strictly related to the proposal here. Generalizing the Downloader isn't strictly required to support remote fetch operations.

Also, instead of calling it Downloader, maybe use "Fetcher" for consistency with the API here?

designs/2020-01-14-remote-downloads.md
```diff
@@ -0,0 +1,589 @@
+---
+created: 2019-04-18
+last updated: 2020-01-20
+status: Approved
+reviewers:
+  - aehlig
+  - buchgr
+title: Remote Downloads
```
ulfjack: Maybe call this "Remote Fetch API"?

Also, I wonder who / whether there should be an example implementation before this can be considered non-experimental? Or put differently, what should be the gate for making this non-experimental?

designs/2020-01-14-remote-downloads.md
sstriker: @ulfjack it may be a different process taking care of pushing.  Even in cases where a [trusted] client is able to push and fetch, different clients are able to benefit from the pushed content.

designs/2020-01-14-remote-downloads.md
sstriker: @ulfjack a Fetch server must know, yes.  In the case of FetchDirectory, the server is asked to return a Directory digest associated with the URI[s]/Qualifier[s].  It must keep track of the types that are associated.  For any set of URI[s]/Qualifier[s] both association types, Blob and Directory, may exist.

designs/2020-01-14-remote-downloads.md
sstriker: @ulfjack Just for pushing, or in general?  If the former, this maintains symmetry with the FetchBlob counterpart.

designs/2020-01-14-remote-downloads.md
sstriker: @ulfjack yes, "Clients **MUST** ensure associated content is available in CAS prior to pushing." per https://github.com/bazelbuild/proposals/pull/160/files#diff-3bdfd5c462602da5c73b9b00c1696223R307

designs/2020-01-14-remote-downloads.md
sstriker: @ulfjack Any http[s]:// endpoint could be a git repository or just serving a file - there's potential for ambiguity there.  Would that suffice as an example?

designs/2020-01-14-remote-downloads.md
sstriker: @ulfjack The client can indicate whether to obtain a Blob or a Directory representation of the content that is being fetched.  For instance in the case of a tarball, the client may want to store the expanded version in CAS.  This allows for:
- pre-expanded storage, saving the download/unpack step on the client
- expanding the concept of builds without bytes, beyond a Tree only content that needs to be accessed locally by the client needs to be retrieved from CAS
- deduplication in case many versions of the tarball exist where the tarball's content is fairly stable

designs/2020-01-14-remote-downloads.md
sstriker: @ulfjack an example could be a client asking for https://github.com/bazelbuild/remote-apis (URI), with scm.branch=master (qualifier).  The server may return an additional qualifier, say scm.commit=b5123b1bb2853393c7b9aa43236db924d7e32d61.

designs/2020-01-14-remote-downloads.md
sstriker: @ulfjack it's up to the server in that case to determine whether what it has cached is fresh enough or whether to fetch a new copy.  Needs some clarifying language.

designs/2020-01-14-remote-downloads.md
jmillikin-stripe: I think this is covered by the first bullet -- too many requests sent to a backend typically manifests as a rate limit denial (assuming a reasonably operated upstream).

designs/2020-01-14-remote-downloads.md
jmillikin-stripe: I'm not sure how you would enforce such policies in the build tool, unless local execution was completely forbidden. In all of the build tools I'm aware of (notably including Bazel), the security boundary must be drawn such that the tool is untrusted.

For discussion of why you can't use a generic proxy for this, see the "Alternatives Considered > HTTP Proxy" section. Most downloads are via HTTPS, for which non-MITM proxies can only control access at the domain level (via SNI).

designs/2020-01-14-remote-downloads.md
jmillikin-stripe: I believe that would be one of the many implementation-defined portions of this spec.

designs/2020-01-14-remote-downloads.md
jmillikin-stripe: Note that the Remote Execution API's content model treats blobs and directories as distinct, but identifies them with the same `Digest`. The server must know ahead of time which type the client is expecting to retrieve. Using different methods avoids a difficult-to-extend enum, and using separate request types is good practice in gRPC to avoid accidentally coupling different code paths.

designs/2020-01-14-remote-downloads.md
jmillikin-stripe: I originally had split this into three documents (remote fetch API, remote push API, Bazel remote downloads) but figured it would be tidier to keep as one.

Would you prefer it be split back up?

designs/2020-01-14-remote-downloads.md
jmillikin-stripe: This part was actually already done: https://github.com/bazelbuild/bazel/pull/10547

designs/2020-01-14-remote-downloads.md
jmillikin-stripe: Done.

designs/2020-01-14-remote-downloads.md
jmillikin-stripe: Done.

designs/2020-01-14-remote-downloads.md
jmillikin-stripe: Done.

designs/2020-01-14-remote-downloads.md
jmillikin-stripe: Done.

designs/2020-01-14-remote-downloads.md
jmillikin-stripe: Done.

designs/2020-01-14-remote-downloads.md
jmillikin-stripe: > Also, I wonder who / whether there should be an example implementation before this can be considered non-experimental? Or put differently, what should be the gate for making this non-experimental?

During in-person discussions at BazelCon, @aehlig and I figured that it'd be best to have it be experimental until the RBE folks had gotten the chance to poke at it a bit. I have no strong feelings about who's making that particular decision.

buchgr: How about user feedback? If enough users find this helpful we can mark it stable. I don't even think we'll need to settle on a number but it will just be obvious at some point.

designs/2020-01-14-remote-downloads.md
ulfjack: Say that explicitly? "If this is not set or set to 0, then the implementation may apply an implementation-defined timeout."

You could also say that there is a maximum/minimum implementation-defined timeout, or that this is mostly advisory.

designs/2020-01-14-remote-downloads.md
ulfjack: I definitely do *not* want it split up into multiple documents. I am just wondering how we expect this to transition to non-experimental.

designs/2020-01-14-remote-downloads.md
ulfjack: What about in general?

designs/2020-01-14-remote-downloads.md
jmillikin-stripe: I can also put it behind a non-experimental flag if you want, `--remote_downloader` or whatever.

Honestly I _do not care_ about the non-Bazel use cases here, and I'm just trying to get this thing pushed through layers of public-facing Google bureaucracy. If the pusher portion is going to be a blocker then I can move it out and let you, Jakob, and Sander hammer out the details there (or not). Please just let the "tell remote service to fetch a blob" feature land so I can stop having builds fail every time golang.org hiccups.

ulfjack: No objection from me. I would just like the doc to be a bit more polished in case I ever have to implement this. :-D

designs/2020-01-14-remote-downloads.md
EricBurnett: 0 is semantically meaningful as-is - it means you'll accept content cached anytime after January 1, 1970, i.e. you don't really care how fresh the content is. (Adding a comment to make this explicit is still reasonable, just noting that it's not strictly necessary).

designs/2020-01-14-remote-downloads.md
EricBurnett: Expiration would not be a good fit for qualifiers - expiration values can change via re-push of the same content, or implicit touch-on-use, or other such things. It's not really a property of the content per-se, and would I think cause confusion to include into Qualifiers. (One could also not _request_ by it, since it's unstable, which is a good sign it's not really a Qualifier).

Digests as qualifiers would work, but I don't think I'd want to. The point of PushBlobRequest is to associate a URL and set of Qualifiers with a Digest, putting the digest *into* the qualifiers makes that relationship unclear. It also means that Digest qualifier is "special" (something the server should interpret, e.g. so it can touch the blob in the CAS; required on Push). This'd also move us back more towards a generalization of inputs/outputs that we reviewed in an earlier proposal and collectively rejected - having clarity around outputs being specifically blob-by-digest and directory-by-digest and nothing more was valued.

designs/2020-01-14-remote-downloads.md
EricBurnett: W.r.t. making it non-experimental, I don't think RBE needs to be the gate on this per se. I'd say once there are two clients and two severs implementing, and no strong requests for change coming from them, I think we can graduate it from experimental?

John: I believe I'm the cause of much of what you call "public-facing Google bureaucracy", for which I do apologise. I have definitely held this proposal to constraints you did not care about for your use-case, including non-bazel support for anything standardized here. I hope that proves to have been worthwhile in the long term, but if not, you can direct your ire at me :).

designs/2020-01-14-remote-downloads.md
jmillikin-stripe: Done.

designs/2020-01-14-remote-downloads.md
jmillikin-stripe: Added a comment to make this explicit.

designs/2020-01-14-remote-downloads.md
jmillikin-stripe: Done.

designs/2020-01-14-remote-downloads.md
jmillikin-stripe: > I would just like the doc to be a bit more polished in case I ever have to implement this. :-D

I'm currently operating under the assumption that I'll be implementing the Bazel side of this protocol, given that there's not currently any Bazel devs working on this area and I really want it.

ulfjack: I can't offer to implement it, but I find the functionality very interesting.

jmillikin-stripe:
@buchgr I think this is now up-to-date with all comments.

buchgr:
I ll give it until Monday before I merge it in case there are any more comments.

designs/2020-01-14-remote-downloads.md
ulfjack: Apologies for reopening this. It sounds like it would be an option if you forbid local execution, so maybe you could mention that here or below in the alternatives considered section? (Technically, if the policy is applied before any local execution, that should be sufficient to guarantee the security boundary?)

ulfjack: I think the interesting scenario is where Bazel doesn't download the remote dependency to the local machine *at all*, but uses this API to get it into the CAS.

designs/2020-01-14-remote-downloads.md
ulfjack: I don't follow. If I have an URL that *I know* was pushed, why would I fetch it? Only to make sure that it stays in the CAS?

designs/2020-01-14-remote-downloads.md
ulfjack: I think it would help to mention the example in the document if it isn't already.

designs/2020-01-14-remote-downloads.md
ulfjack: Ok, thanks.

designs/2020-01-14-remote-downloads.md
ulfjack: I don't find that quote with Ctrl-f in this document.

designs/2020-01-14-remote-downloads.md
ulfjack: I'd like to have an example in the doc. Yes, disambiguating between git repositories and files seems like a good example.

ulfjack:
Sorry for the delay. I added a couple more comments, but I'm also ok with merging this as-is.

I would like to see a setup where Bazel doesn't read or write any files from/to local disk in the future, and everything happens remotely, and I think this is a step towards that.

sstriker:
No worries.

The setup of having no files on local disk is one of the motivations for
the API.

designs/2020-01-14-remote-downloads.md
jmillikin-stripe: Bazel still needs to download the file so that it can be validated against the checksum. Otherwise a malicious remote downloader could inject arbitrary content into the action cache.

Completely forbidding local execution would break enormous parts of the Bazel ecosystem, including Bazel's C++ toolchain configuration and every repository rule that needs to execute OS-specific setup. If some future Bazel v100.x supports this mode of operation then we can revisit.

designs/2020-01-14-remote-downloads.md
jmillikin-stripe: I believe Sander wants to have some trusted external thing push (for example) `urn:uuid:5EE2EC4C-DB23-4ED3-BFDA-6236B429536A` with some content, then have BuildStream reference the content by that identifier to get a `Digest`. Without the intermediate step you'd have to identify things by `Digest` directly, which would be non-portable between storage implementations (different digest algos in use).

jmillikin-stripe: I believe this is already covered by the existing text:

> URNs may be used as opaque content identifiers, for instance by using the `uuid`
> namespace: `urn:uuid:f81d4fae-7dec-11d0-a765-00a0c91e6bf6`. This is most
> applicable to content that is injected into the CAS via the Push API, where the
> URN serves as an agreed-upon key, but carries no other inherent meaning.

designs/2020-01-14-remote-downloads.md
EricBurnett: **All:** Worth noting, this API is intended to cover a variety of usage patterns, of which "trusted downloader", "untrusted downloader", "trusted build tool", "untrusted build tool", etc all can benefit. If anything, this Background is overly strong about the build tool not being the one to do the downloading - I fully expect bazel to end up implementing Push, for example, alongside UpdateActionResult for use-cases benefitting from optimistic best-effort caching among mutually-trusted builders.

**John:** Assuming this topic is still open and seeking agreement, I'd recommend replacing the sentence "_There are three important reasons why the build tool should not be the thing doing the actual downloading"_ with something to the effect of "_Use of this API can benefit a variety of use-cases:_" ?

(I don't personally care - I'm also fine committing it as-is. But I'm guessing this change might make the text more generally agreeable).

**Ulf:** Adding "alternatives" I think puts too much emphasis on this use-case - I hope we can all agree that it's reasonable for some parties to not want to trust the client tool, in which case this API can be useful in asserting some security boundaries, and leave it at that? If, as recommended above, this is reframed as examples moreso than applicable to everyone, that should hopefully make it clearer that not _everyone_ must approach things in exactly this way. Certainly use of this API is not mutually exclusive with trusting bazel to enforce a security boundary, and in many cases that would actually be complementary.

**John:** to your last point, if the action cache is being trusted then it seems reasonable that the download service may also be trusted? I.e. if bazel is going to trust contents it retrieves from the Action Cache to have been run in the right environment and whatnot, it'd make sense to me that it should also trust a downloader service to have fetched the right remote resources, and certainly to do simple things like correctly unpack a known (checksummed) tarball.

(I can also see use-cases for untrusted downloaders, especially with arbitrary servers inserting contents via Push, just pointing out that there are reasonable scenarios where you *would* want to trust the service and not re-hash everything yourself.)

jmillikin-stripe: I honestly can't imagine any reason that Bazel would use `Pusher` -- the code I've written on the Bazel side only calls `/Fetch.FetchBlob`, and given the lack of momentum on this feature from the Bazel side there doesn't seem to be another strongly requested use case here.

Services that execute build actions and services that fetch data from the Internet are often mutually untrusted, and may be run by separate teams.

If Bazel were to be treated as trusted then it would need a lot of changes (like the ability to block all non-execroot file access, block `repository_ctx.execute()`, block workspace status scripts, ...). I've also heard the Bazel devs state that Bazel is not designed to operate in an untrusted source tree.

I understand that you envision this as some sort of all-purpose API but that is very much not what I need, it's not what I'm going to send a PR to implement, and I don't know of anyone on the Bazel side who wants that sort of API.

jmillikin-stripe:
Bazel client implementation: https://github.com/bazelbuild/bazel/pull/10622

designs/2020-01-14-remote-downloads.md
ulfjack: The invocation policy infrastructure allows blocking workspace status scripts as well as forbidding local execution. Google uses a version of Bazel that doesn't support external repositories. Local file access is only security-critical if there are world-readable security-relevant files on the local disk.

I would go so far as to say it can be run that way, and we need to keep it that way, generally speaking. If there are Google employees claiming that Bazel isn't designed to do that, then they should come talk to me. :-D

buchgr:
@jmillikin-stripe please rebase and I ll merge.

sstriker:
> @jmillikin-stripe please rebase and I ll merge.

@buchgr don't want to wait for bazelbuild/remote-apis#112 to land?  [Vote on the name](https://groups.google.com/d/msg/remote-execution-apis/kTcjXrijWpU/Z2z-atEHEgAJ) is open for another 10h or so.

buchgr:
@sstriker it seems like "Asset" is the name. I'll wait for your PR to land.

designs/2020-01-14-remote-downloads.md
EricBurnett: > I honestly can't imagine any reason that Bazel would use Pusher

Same reason bazel supports `UpdateActionResult` today - a fairly lightweight way for multiple builders to pool effort. With `UpdateActionResult`, instead of each independently executing actions, they can put them in a central cache and reuse the results from any of their peers. Not perfectly secure, but plenty good enough for some use-cases. Likewise, instead of each independently downloading inputs from origin, the first to download a given asset could do so and push it into a shared cache, for the rest to pull from instead. This is essentially what the repository cache does, except it'd be shareable across machines without needing a shared _filesystem_.

To be clear, certainly I don't expect you to implement that, or that it'll necessarily land right away from anyone. But I have definitely heard requests for such a feature, and do expect _someone_ to eventually.

jmillikin-stripe:
Rebased to current master.

I've not done a `s/Fetcher/Fetch/` or similar updates to sync with https://github.com/bazelbuild/remote-apis/pull/112 because it's a hassle to check the text for mismatches, and the canonical definition of the protocol is always going to be the `.proto` files.

jmillikin-stripe:
Rebased again. @buchgr is this good to merge? The .proto specification has already merged into bazelbuild/remote-apis.

buchgr:
@jmillikin-stripe almost. Can you please add a link to the API specification? https://github.com/bazelbuild/remote-apis/blob/master/build/bazel/remote/asset/v1/remote_asset.proto

jmillikin-stripe:
> Can you please add a link to the API specification?

Done.

PR #171, opened 2020-02-11
Re-move starlark project files to dropped.

robbertvanginkel:
This was moved to dropped in #165, but seems to have been accidentally re-added when merging 4179d9f219a48872b72674c878168644ed77be62.


jin:
Thank you!

PR #174, opened 2020-02-28
Under review: "Measuring Configuration Overhead"

gregestren:
In accordance with design review policy, this is just to put the doc in the registry, not the actual review.


sdtwigg:
LGTM

PR #179, opened 2020-04-17
Add Dependency Discovery Proposal to Drafts

stevebillings:


stevebillings:
@aiuto Here is the PR that we've been discussing by email, for the Dependency Discovery proposal (PRD).

aiuto:
Thanks for starting up this effort Steve.

PR #190, opened 2020-05-26
Mark Starlark Directory Expansion in Args MapEach as Approved.

haileiyu:
Also updated the 'author' field to tomlu@.


lberki:
Erm. We managed to not notice this pull request, but this apparently got implemented since then and I moved the design doc to the "Implemented" state so I'll close it.

PR #194, opened 2020-07-17
Update README.md

WarkahScott:
Added shorthand flag mapping proposal


gregestren:
Let's put this in the "Under Review" section, like at https://github.com/bazelbuild/proposals/blob/8115a567ea088f338ab6f6c4f54d9547d1217c33/README.md#under-review?

WarkahScott:
I didn't notice it was in the wrong section, thanks for the fix.

PR #196, opened 2020-09-29
Add proposal "Adding a constraint to default target platform".

comius:


katre (approved):
Approving PR to start the review process.

PR #197, opened 2020-09-29
Replace embedded JDK with JDK from remote repository.

comius:


lberki (approved):
I got around reading the discussion thread and was wondering why this document still refers to the embedded JDK. Turns out, I should have been more diligent with code reviews!

PR #223, opened 2021-04-28
Remote Output Service: Mention read-only output paths

EdSchouten:
@alexjski made a fair point on the mailing list thread last week that it
may make sense to make the output path completely read-only. I agree,
though it would break unsandboxed actions.

For now, let's stick to keeping the output path writable, but at least
add a section to the proposal on what needs to be done to make it
read-only in the future.


EdSchouten:
Cc @alexjski @janakdr

lberki (approved):

PR #242, opened 2022-01-21
Add design proposal for optional toolchains

katre:
This is just to begin the design discussion, not to set the proposal as complete.


gregestren (approved):
LGTM in terms of adding the proposal (not doing design review in this PR)

PR #262, opened 2022-06-07
Add proposal for credential helper

Yannic:


lberki (approved):

Yannic:
@coeuvre you seem like a good fit as lead reviewer for this, PTAL.

@meteorcloudy PTAL as well.

lberki:
Almost! We assigned @tjgq (for remote execution) and @Wyverald (for repository fetching) to the review so I sent out https://github.com/bazelbuild/proposals/pull/263 to change the reviewers.

@bergsieker might have something useful to say, too so I added him as a third reviewer, in case he has the time.

designs/2022-06-07-bazel-credential-helpers.md
```diff
@@ -0,0 +1,100 @@
...
+services. Many other applications have solved similar problems in the past.
+
+
+# Proposal
+
+We propose to allow Bazel to retrieve credentials from an (external) helper
+binary. This standard approach is used by, e.g., `git` or `docker`, both of
+which face similar problems for connecting to remote services.
+
+A credential helper can be any program that can read values from standard input.
```
bergsieker, line 51: How will this handle situations where a user needs multiple credentials? For example, it's reasonable to pull source from a private repository, then use a GCP-based remote execution system. My understanding of the current proposal is that the credential helper would have to be a wrapper that would inspect the URL and provide the appropriate credential. As an alternative, what about passing URL:helper pairs to bazel so that the URL-inspection code could be implemented directly by Bazel?

designs/2022-06-07-bazel-credential-helpers.md
```diff
@@ -0,0 +1,100 @@
...
+  message in its standard error that Bazel will print.
+
+Whenever connecting to a remote server, Bazel will execute the credential helper
+to retrieve credentials. Since there are usually many connections to the same
+service (e.g., in the form of RPCs against a `Remote Cache`), Bazel may
+(optionally) cache credentials to avoid running the credential helper binary too
+often.
+
+Since Bazel wouldn't know which credential helper to call, users must
+explicitly configure Bazel by passing `--credential_helper`, which can either
```
bergsieker, line 80: Assuming this flag eventually supersedes the existing credential-related flags, how would the existing functionality be maintained? Would users always need to provide their own credential helper, or would Bazel bundle (some of) the existing functionality into an off-the-shelf helper for common scenarios?

Yannic: I could see Bazel bundling helpers for `.netrc` or `GCP` in the future, but deprecating the existing credential-related flags is out of scope for this proposal.

designs/2022-06-07-bazel-credential-helpers.md
Yannic: My original idea was to delegate that to the credential helper, e.g., by using a wrapper that reads from, e.g., a config file. An alternative could be to do `--credential_helper=<domain>=<helper>` (where `<domain>` may support wildcard domain names, so `*.example.com` would match `example.com`, `www.example.com`, `foo.example.com`, `...`.

I'll update the proposal.

PR #264, opened 2022-06-10
Handle multiple credential helpers for domains and allow workspace-relative helpers

Yannic:


designs/2022-06-07-bazel-credential-helpers.md
```diff
@@ -80,11 +80,33 @@ often.
...
-be an absolute path, or a string. In case the value is a string (`foo`), Bazel
-will look for an executable with the name `foo-bazel-credential-helper` on the
-path. Since the credential helper must be available very early in the build
-(i.e., during the loading phase for fetching external repositories), it cannot
-be a label to a target.
+be an absolute path, or a string. In case the value is a string (e.g.,
+`foo-bazel-credential-helper`), Bazel will look for an executable with the
+provided name on the path. As a special case of absolute paths, the value may
+start with `%workspace%` (e.g., `%workspace%/foo/bar/baz.sh`), in which case
+`%workspace%` is replaced by the absolute path to the workspace. Since the
```
adam-azarchs, line 87: For consistency, I think the semantics of this should be the same as [--workspace_status_command](https://docs.bazel.build/versions/main/user-manual.html#flag--workspace_status_command).  Granted those are not very well documented, but as far as I can tell, it's behavior is the same as how a shell would interpret it when executing from the workspace directory.  That is, absolute paths are treated as absolute paths, paths containing a `/` are interpreted as relative paths (relative to the repo root) and paths which do not contain a `/` are searched for in the `PATH`.

While there might be many use cases for a syntax like `%workspace%` in `.bazelrc`, at present no command that I know of honors any kind of variable expansion like that (which can be unfortunate at times).  If such a thing were to be introduced, I think that should be done in a way that is applied consistently across all flags.

designs/2022-06-07-bazel-credential-helpers.md
```diff
@@ -80,11 +80,33 @@ often.
...
+credential helper must be available very early in the build (i.e., during the
+loading phase for fetching external repositories), it cannot be a label to a
+target.
+
+When executing the credential helper, Bazel will always set the path to the
+workspace as working directory (i.e., `$(pwd)` will be `%workspace%`).
+
+Since users may need to use multiple credential helpers (e.g., one for fetching
+external repositories and another one for connecting to a `Remote Cache`), users
+may optionally specify the (wildcard) DNS name for which the credential helper
```
adam-azarchs, line 97: There are some important use cases for being able to specify things based on a full path prefix rather than simply DNS name.  For example, one might have a bunch of artifacts to download from `s3-*.amazonaws.com`, some of which might be public buckets but others of which might not be, and might wish to only send credentials for the non-public buckets.  Another obvious example would be `github.com`, where one might want to send credentials for access to certain orgs but not others.  Yes, the helper script can do that additional filtering, but there's various reasons why one might prefer it not get that far, including security (not wanting to accidentally send credentials to the wrong place due to a typo in the helper script) and performance (not wanting to make bazel invoke the credential helper unnecessarily).

designs/2022-06-07-bazel-credential-helpers.md
```diff
@@ -80,11 +80,33 @@ often.
...
+should be used as prefix to the path of the binary, seperated by
+`%path_separator%` (i.e., `:` on Linux/macOS and `;` on Windows). In case there
```
adam-azarchs, line 99: For consistency with other flags which take pairs of arguments like this (e.g. [`--override_repository`](https://docs.bazel.build/versions/main/command-line-reference.html#flag--override_repository), [`--strategy`](https://docs.bazel.build/versions/main/command-line-reference.html#flag--strategy), [`--starlark:file`](https://docs.bazel.build/versions/main/command-line-reference.html#flag--starlark:file)), I would suggest using `=` as the separator.  Yes, technically `=` can be part of a filename but so can `:` or `;`.  At the very least it would be better to keep it consistent between operating systems (especially since `:` is not allowed in file names in Windows but `;` is).

designs/2022-06-07-bazel-credential-helpers.md
```diff
@@ -80,11 +80,33 @@ often.
...
+are multiple credential helpers specified, Bazel will use the most specific
+match to fetch credentials. For example, assuming a user passes
+`--credential_heler=foo --credential_helper=*.example.com:bar --credential_helper=example.com:baz`,
```
tjgq, line 102: Typo in the first `--credential_helper`.

designs/2022-06-07-bazel-credential-helpers.md
tjgq: +1 for `=`:

* I don't think this should be platform-dependent; it makes it more cumbersome to reuse the same `.bazelrc` across platforms.
* I can envision a future extension where the left hand side may include a port number (e.g. `example.com:8080`) so `:` might be a bad choice.

We can also clarify that we split on the first occurrence, so the right hand-side is still allowed to contain the separator.

designs/2022-06-07-bazel-credential-helpers.md
tjgq: https://github.com/bazelbuild/bazel/issues/14861 seems relevant here. Is that a strong enough argument for `%workspace`? (I do want to avoid unnecessarily expanding the scope of this proposal, though; even if we agree that `%workspace` is a good idea for other flags, that should be pursued separately.)

Regardless of whether `%workspace%` is supported, I think the proposal should still clarify how paths containing a `/` are handled, whether absolute or relative.

designs/2022-06-07-bazel-credential-helpers.md
```diff
@@ -80,11 +80,33 @@ often.
...
+Bazel will run `baz` to get credentials for `example.com`, `bar` for fetching
+credentials for `a.example.com` or `x.y.z.example.com`, and `foo` for fetching
+credentials for any other DNS name. In case there are multiple credential
+helpers specified for the same (wildcard) DNS name, the later entry overrides
```
tjgq, line 106: Can you precisely define what "wildcard DNS name" means?

For example, with the current wording it's not obvious whether `s3-*.amazonaws.com` would be valid (see comment above). I'm also assuming we only want to support `*` as a "zero or more characters" placeholder (i.e., no `?` or more advanced regex-like operators) but let's make it clear.

tjgq (changes requested):

aiuto (commented):
I don't have time for a detailed review. Just a few thoughts.
1. Yes. The need for this is real. Thanks for proposing.
2. The reviewers should include people from the info-sec world, rathter than just Bazel people.
3. I think we should go much further.  It should be URL downloader.

On the third point, I think we could make Bazel safer in many respects if all downloads were done by a separate binary. Bazel would come with a reference implementation, but a user should be able to create their own. This would increase transparency of network I/O code, and allow local custimization.  The downloaders would handle credential setting (either by embedding the helper here, or using the protocol defined here to talk to it), as well as more interesting tasks like rewriting URLs. For example, a user in APAC should not have to rely on a mirror in the US. They should be able to read through their own local cache.

I'll admit I have not thought through how that interacts with RPC calls. Presumably we would not have to rewrite URLs for RPC calls because it is the user who gets to put the endpoint in the .bazelrc file. It is dependency downloading where a user is most likely to not want to use/trust the URL provided in a deps.bzl or bzlmod file and fetch from an alternate location.

designs/2022-06-07-bazel-credential-helpers.md
Yannic: I'm aware that `%workspace%` is currently not honored by any flags (but it is by other parts of Bazel, so it's not a completely new concept). TBH, I wish it was, it would make, e.g., the semantics of `--workspace_status_command` so much clearer.

I'd like to stick with `%workspace%` and then simply forbid non-absolute paths containing any `/`s because the semantics of relative paths aren't uncontroversial (relative to the workspace, to the current working directory of the user invoking Bazel, relative to the `.bazelrc` it's set in?). I'll write a separate proposal to allow `%workspace%` in `--workspace_status_command` & friends.

designs/2022-06-07-bazel-credential-helpers.md
Yannic: Noted (and agreed). However, I would like to start with the simple DNS name matching and extend if needed. Doing full-path-prefix matching makes things much more complicated and I'm not convinced that complexity should be in Bazel vs. a credential helper (which can have its own config file, not being restricted to what we can do with Bazel flags). I think that the DNS name matching I'm proposing here will allow us to extend it to full-path-prefix matching in the future should we need it.

designs/2022-06-07-bazel-credential-helpers.md
Yannic: ack

designs/2022-06-07-bazel-credential-helpers.md
Yannic: fixed

designs/2022-06-07-bazel-credential-helpers.md
Yannic: Clarified that we only allow wildcards that match all subdomains (i.e., only `*.example.com`), and that regexes are not allowed (for now).

Yannic:
@aiuto Thanks for your thoughts! I think rewriting URLs (e.g., to be geographically closer or an internal mirror of something on GitHub) is probably something we should look into as well. I think it's out of scope for this proposal though. I think allowing users to customize the downloader to, e.g., only allow fetching from a specific company-internal GCS/S3 bucket, is a nice idea as well. Unfortunately, similarly out of scope for this proposal. Let's do one step at a time :).

Adding some info-sec people to the review is fine with me.

aiuto:
I agree with both of you on doing credential helpers better now as a first goal.

@lberki   I meant reviews people from security teams at Google and other interested companies. They are more attuned to finding the potential for credential leakage than we are.

adam-azarchs:
There's definitely some infosec implications, in particular that cloning a repo and then running `bazel fetch` would involve arbitrary code execution (something golang has gone out of its way to prevent).  Of course repository rules already permit arbitrary code execution.  Allowing credential helpers to do this as well _might_ make attacks like that a little harder to detect, however, particularly if we allow workspace-relative credential helper scripts to be specified in `.bazelrc`.  Personally I don't think it's worth worrying about that unless something is also done to address the issue for repository rules first (which is probably a good idea but also entirely out of scope for this proposal).

Infosec people would probably also say that the proposal should specify that credential helpers will never be invoked for unencrypted connections (e.g. `http` without the `s`).

> On the third point, I think we could make Bazel safer in many respects if all downloads were done by a separate binary. Bazel would come with a reference implementation, but a user should be able to create their own. This would increase transparency of network I/O code, and allow local custimization. The downloaders would handle credential setting (either by embedding the helper here, or using the protocol defined here to talk to it), as well as more interesting tasks like rewriting URLs.

I agree this would add a lot of power, potentially.  However I think it's important to keep in mind that the dominant use cases for what we're talking about are really quite simple things like authenticating to private github repositories and s3 buckets, and for smaller projects/organizations needing to implement an entire downloader binary just to provide an auth token would be a a disproportionate amount of effort.  I think it's an interesting idea to pursue, but keeping it separate from providing a [relatively convenient] way to provide credentials for standard https downloads.

Yannic:
> Infosec people would probably also say that the proposal should specify that credential helpers will never be invoked for unencrypted connections (e.g. http without the s).

Restricting helpers to TLS seems fine with me

designs/2022-06-07-bazel-credential-helpers.md
```diff
@@ -69,22 +69,68 @@ command to execute. For now, there will only be one valid value:
...
-(i.e., during the loading phase for fetching external repositories), it cannot
-be a label to a target.
+explicitly configure Bazel by passing `--credential_helper=<path>`, where
+`<path>` can either be an absolute path, or the name of an executable. In case
+`<path>` is the name of an executable (e.g., `foo-bazel-credential-helper`),
+Bazel will look for an executable with the provided name on the `PATH`. The name
+of the executable must not contain any path seperators (`/`, and `\` on
+Windows). As a special case of absolute paths, the value may start with
+`%workspace%` (e.g., `%workspace%/foo/bar/baz.sh`), in which case `%workspace%`
+is replaced by the absolute path to the workspace. Since the credential helper
```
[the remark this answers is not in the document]
Yannic: I think `PATH` lookup is unfortunately not something we can get rid of :/. Think of the cases where a (commercial) remote cache/execution service distributing their credential helper to authenticate to their service. Yes, users could check that into the repo, but I expect people are hesitant to check in a binary, and forcing everyone to create a wrapper that then does `PATH` lookup doesn't sound compelling either. There's also the problem of having multiple credential helper binaries for different platforms (Linux vs. macOS vs Windows, x86 vs arm, ...). As much as I'd like to avoid `PATH` lookup myself, I don't think we can while making this feature useful to most people. If someone doesn't want `PATH` lookup for their case, they can specify an absolute path or a workspace-relative path instead.

Yannic:
> Another thing that occurs to me from an infosec point of view is handling of redirects.

I would expect that Bazel doesn't use credentials on redirect, and instead invoke the credential helper for the destination domain again.

designs/2022-06-07-bazel-credential-helpers.md
```diff
@@ -69,22 +69,68 @@ command to execute. For now, there will only be one valid value:
...
+helper, seperated by the left-most occurence of `=` (i.e.,
+`--credential_helper=<pattern>=<path>`, where `<path>` is as defined above).
+`<pattern>` can either be a DNS name matching a single domain (e.g.,
+`example.com`), or a wildcard DNS name matching the domain itself and all
+subdomains by prefixing the DNS name with `*.` (e.g., `*.example.com` would
+match `example.com`, `foo.example.com`, and so on). Wildcards that match only a
+subset of all possible subdomains, or regular expressions, are not supported.
+
+In case there are multiple credential helpers specified, Bazel
+will use the most specific match to fetch credentials. For example, assuming a
```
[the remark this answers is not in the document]
adam-azarchs: I would assume that the default behavior if no URL matches would be to not use a credential helper at all, and thus not send any authentication headers of any kind.  That's the secure option, and also improves performance for the common case where one is downloading something from a URL that doesn't require authentication.  But you're right that should be made explicit.

designs/2022-06-07-bazel-credential-helpers.md
```diff
@@ -69,22 +69,68 @@ command to execute. For now, there will only be one valid value:
...
+external repositories and another one for connecting to a `Remote Cache`), users
+may optionally specify a pattern to match which addresses to use the credential
+helper for by passing `<pattern>` as prefix to the path of the credential
+helper, seperated by the left-most occurence of `=` (i.e.,
+`--credential_helper=<pattern>=<path>`, where `<path>` is as defined above).
+`<pattern>` can either be a DNS name matching a single domain (e.g.,
+`example.com`), or a wildcard DNS name matching the domain itself and all
+subdomains by prefixing the DNS name with `*.` (e.g., `*.example.com` would
+match `example.com`, `foo.example.com`, and so on). Wildcards that match only a
+subset of all possible subdomains, or regular expressions, are not supported.
```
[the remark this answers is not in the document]
adam-azarchs: I agree that in general subdomains should match implicitly, as that's a common use case (e.g. `google.com` or `github.com`).  When I'd originally suggested having wildcard support I was thinking of the example of `s3-*.amazonaws.com` which is not all subdomains for `amazonaws.com`.  Though I suppose in that case it's not a serious issue if other domains match (it's mostly just making the match more self-documenting).

designs/2022-06-07-bazel-credential-helpers.md
```diff
@@ -69,22 +69,68 @@ command to execute. For now, there will only be one valid value:
...
 often.
 
+Credential helpers cannot rely on `std{in,out,err}` to communicate with the
+user (i.e., Bazel will not show any output of the credential helper to the user
+or forward any inputs from the user to the crednetial helper). Workflows that
+require user-interaction (e.g., for 2FA, e.g., with a Yubikey) will need to
+notify the user about the required interaction through another channel (e.g.,
+using a OS dialog or opening a browser). Bazel will, however, print a message
+with at least the path of the credential helper as part of its normal output
+whenever running a credential helper (at `INFO` level).
```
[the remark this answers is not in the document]
adam-azarchs: I think "should not" is the right keyword there, in the [RFC 2119](https://www.ietf.org/rfc/rfc2119.txt) sense of the word: it's a good idea for credential helpers to not require user interaction, but there isn't really a way for bazel to enforce that so it wouldn't be "must not".  My main point was that bazel should definitely enforce a timeout on credential helpers (particularly necessary on CI systems where a helper which decided it needed user interaction would be waiting indefinitely).

designs/2022-06-07-bazel-credential-helpers.md
```diff
@@ -69,22 +69,68 @@ command to execute. For now, there will only be one valid value:
...
+  If the credential helper exists with any exit code `!= 0`, or if its output
+  does not follow the protocol (e.g., is invalid JSON), Bazel will abort the
+  operation with an appropriate `FailureDetail` indicating the reason (i.e.,
+  the build fails). A credential helper may provide a human-readable error
   message in its standard error that Bazel will print.
 
+  TODO(yannic): Specify the appropriate `FailureDetail`. Is there an existing
+  one we can use or do we need to define a new one?
+
 Whenever connecting to a remote server, Bazel will execute the credential helper
```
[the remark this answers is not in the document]
Yannic: I think the most severe case is for remote caching execution, where invoking the credential helper for every RPC is clearly unacceptable (IIRC, we've seen a case where remote builds became slow because of _something something about getting and/or validating credentials_). Hence the sentence to allow Bazel to cache credentials internally. I'm definitely less worried about the overhead invoking a helper has for fetching external repositories.

designs/2022-06-07-bazel-credential-helpers.md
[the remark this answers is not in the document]
adam-azarchs: Ok, but still should still be reworded to make it clear that absolute paths are OK, and also to make it clear which `PATH` will be used (path from the shell where `bazel` was launched vs. `--repo_env` or `--action_env`).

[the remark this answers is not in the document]
Yannic: I planing to do "whatever `--workspace_status_command` does". I think that's using it from the shell Bazel runs in, but I may be wrong. We should absolutely not use `--action_env`.

(in addition to `PATH`, this also applies to which env vars are available to the helper. I'm leaning towards the ones from the shell, but I could be convinced otherwise). In any case, there will be people arguing for every one of these options, so we have to make a call what to do. I'm not convinced we need to specify this in the proposal as long as we document it in the end.

designs/2022-06-07-bazel-credential-helpers.md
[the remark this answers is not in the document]
tjgq: > If bazel is allowed to cache credentials internally then it needs to be specified what the credential helper is allowed to depend on.

This is a great point.

The simplest solution is to declare that the validity scope of the credentials matches the pattern in the `--credential_helpers` flag. This means it won't be possible to return different credentials based on URL components other than the host, but I'm wondering if there are any real-world examples where that would be useful.

A more ambitious alternative would be for the helper response to specify the set of URLs the returned credentials are valid for. For example, if you request credentials for `foo.example.com/path/to/resource`, the helper could return `url: "*.example.com"` to indicate that the credentials are actually valid for the entire `*.example.com` domain. But this adds quite a bit of complexity, especially if the returned pattern must accommodate wildcards on URL components other than the host.

> The helper also needs to be able to specify a TTL on the credentials it returns so that bazel knows when to refresh them.

We definitely need to support the case where the returned credentials has no attached TTL (in the sense that it's unknown, not in the sense that it's infinite). On the other hand, knowing the TTL doesn't really avoid the need for Bazel to react to an authentication failure by requesting credentials again, because an access token can be revoked before its TTL expires. So I suspect including the TTL in the response might be unnecessary complexity.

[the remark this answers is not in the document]
adam-azarchs: > The simplest solution is to declare that the validity scope of the credentials matches the pattern in the `--credential_helpers` flag. This means it won't be possible to return different credentials based on URL components other than the host, but I'm wondering if there are any real-world examples where that would be useful.

The obvious example would be `github.com/my_org` vs. other github orgs on which one doesn't have any special permissions.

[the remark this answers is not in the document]
Yannic: My plan is to start with caching credentials on URL, and invoke the helper again (or another helper) if anything is different. Yes, that likely results in running the helper more often than necessary (esp. for fetching repositories where basically all URLs are unique), but I'd strongly prefer to not have that complexity right from the start. The helper protocol is designed in a way that easily allows such extensions in the future.

> The helper also needs to be able to specify a TTL on the credentials it returns so that bazel knows when to refresh them.

I agree with @tjgq that we have to handle the case where a download or RPC fails because the credentials expired independently of whether the TTL is reached or if there's a TTL at all. For simplicity, let's defer this for now. Again, the protocol will allow us to add a TTL in the future should we choose to need it.

> The simplest solution is to declare that the validity scope of the credentials matches the pattern in the --credential_helpers flag. This means it won't be possible to return different credentials based on URL components other than the host, but I'm wondering if there are any real-world examples where that would be useful.

Another example is GCS/S3, where it's possible to access buckets as `storage.googleapis.com/<bucket-name>`, where you may need different credentials depending on the bucket.

Yannic:
I think I addresses all comments here. Could we merge this iteration in? @tjgq @lberki

tjgq (approved):

PR #266, opened 2022-06-30
Use boolean instead of string parameter for repeatable Starlark flags

fmeum:
@aranguyen @gregestren This is the update to the proposal following the result of our discussion.

### DIFF
--- a/designs/2022-03-06-representation-of-repeatable-starlark-flags.md
+++ b/designs/2022-03-06-representation-of-repeatable-starlark-flags.md
@@ -61,16 +61,14 @@ This approach to repeatable flags has the following drawbacks:
 Repeatable string flags could instead be represented as `string_list` build settings that only
 differ from the default one in the way usages of the flags on the command-line are parsed.
 
-Specifically, a new named parameter `flag_style` can be added to `config.string_list` with the
+Specifically, a new named parameter `repeatable` can be added to `config.string_list` with the
 following behavior:
 
-* The type of the parameter is `String`.
-* The default value of the parameter is `None` (i.e., the parameter is optional).
-* The allowed values for the parameter are `None`, `"comma"`, and `"repeated"`.
-* It is an error to set `flag_style` to a value other than `None` if `flag` is not set to `True`.
-* If `flag` is `True` and `flag_style` is either `None` or `"comma"`, the `string_list` setting
-  behaves as it does now.
-* If `flag` is `True` and `flag_style` is `"repeated"`, the `string_list` setting behaves as it does
+* The type of the parameter is `boolean`.
+* The default value of the parameter is `False` (i.e., the parameter is optional).
+* It is an error to set `repeatable` to a value other than `False` if `flag` is not set to `True`.
+* If `flag` is `True` and `repeatable` is set to `False`, the `string_list` setting behaves as it does now.
+* If `flag` is `True` and `repeatable` is set to `True`, the `string_list` setting behaves as it does
   now with the exception that the value of the setting set on the command line is a list containing
   the unmodified values of each individual occurrence of the flag in the order they appear. In
   particular, the values are *not* split on a comma. This parsing behavior agrees with that of a
@@ -82,7 +80,7 @@ With this new parameter, the example from the previous section could look as fol
 # rules_cc/cc/internal/config.bzl
 repeatable_string_flag = rule(
     implementation = _impl,
-    build_setting = config.string_list(flag = True, flag_style = "repeated")
+    build_setting = config.string_list(flag = True, repeatable = True)
 )
 
 # rules_cc/cc/BUILD.bazel
@@ -110,9 +108,10 @@ Starlark semantics flag defaulting to `true` could be added and used specified i
 
 ## Alternatives considered
 
-Instead of a `String` parameter `flag_style`, the new parameter of `config.string_list` could be a
-`boolean` named `repeated` or `allow_multiple` that enables the new flag behavior. However, this
-would make it less clear that the individual flag values are not split on commas.
+Instead of a `boolean` parameter `repeated`, the new parameter of `config.string_list` could be a
+`String` argument that allows to specify the name of a "flag passing style". This would allow for
+the introduction of flag styles other than comma-separated and repeated and potentially be more
+self-documenting, but also complicates the API without a clear current need.
 
 # Backward-compatibility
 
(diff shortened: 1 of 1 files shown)


designs/2022-03-06-representation-of-repeatable-starlark-flags.md
```diff
@@ -61,16 +61,14 @@ This approach to repeatable flags has the following drawbacks:
...
-* The type of the parameter is `String`.
-* The default value of the parameter is `None` (i.e., the parameter is optional).
-* The allowed values for the parameter are `None`, `"comma"`, and `"repeated"`.
-* It is an error to set `flag_style` to a value other than `None` if `flag` is not set to `True`.
-* If `flag` is `True` and `flag_style` is either `None` or `"comma"`, the `string_list` setting
-  behaves as it does now.
-* If `flag` is `True` and `flag_style` is `"repeated"`, the `string_list` setting behaves as it does
+* The type of the parameter is `boolean`.
+* The default value of the parameter is `False` (i.e., the parameter is optional).
+* It is an error to set `repeatable` to a value other than `False` if `flag` is not set to `True`.
```
aranguyen, line 69: nice edge case handling here! Thanks! What about when users set both `allowMultiple` and `repeatable`? Do you think it would make sense to add a check so that we do not allow users to set both?

designs/2022-03-06-representation-of-repeatable-starlark-flags.md
```diff
@@ -82,7 +80,7 @@ With this new parameter, the example from the previous section could look as fol
 # rules_cc/cc/internal/config.bzl
 repeatable_string_flag = rule(
     implementation = _impl,
-    build_setting = config.string_list(flag = True, flag_style = "repeated")
+    build_setting = config.string_list(flag = True, repeatable = True)
```
aranguyen, line 83: Eventually, after everyone has migrated to the new representation, are we planning on either remove `allowMultiple` or `repeatable`?

designs/2022-03-06-representation-of-repeatable-starlark-flags.md
```diff
@@ -61,16 +61,14 @@ This approach to repeatable flags has the following drawbacks:
...
-* If `flag` is `True` and `flag_style` is either `None` or `"comma"`, the `string_list` setting
-  behaves as it does now.
-* If `flag` is `True` and `flag_style` is `"repeated"`, the `string_list` setting behaves as it does
+* The type of the parameter is `boolean`.
+* The default value of the parameter is `False` (i.e., the parameter is optional).
+* It is an error to set `repeatable` to a value other than `False` if `flag` is not set to `True`.
+* If `flag` is `True` and `repeatable` is set to `False`, the `string_list` setting behaves as it does now.
+* If `flag` is `True` and `repeatable` is set to `True`, the `string_list` setting behaves as it does
   now with the exception that the value of the setting set on the command line is a list containing
   the unmodified values of each individual occurrence of the flag in the order they appear. In
```
aranguyen, line 73: So how does this look on the command line? Would this just be `repeatable_flag=[val1, val2, val3]` vs the current behavior`allowMulitple_flag=val1 allowMulitple_flag=val2 allowMulitple_flag=val3`? 

This is optional. I like how you had the code block example below for the starlark definition. If you could do something similar for this part, it would be really great. Thank you!

designs/2022-03-06-representation-of-repeatable-starlark-flags.md
```diff
@@ -61,16 +61,14 @@ This approach to repeatable flags has the following drawbacks:
...
   particular, the values are *not* split on a comma. This parsing behavior agrees with that of a
```
aranguyen, line 74: ok maybe my interpretation the representation on the command line in the comment above `repeatable_flag=[val1, val2, val3]` is in correct?

designs/2022-03-06-representation-of-repeatable-starlark-flags.md
fmeum: `repeatable` is a new parameter on `config.string_list`, whereas `allow_multiple` is an existing parameter on `config.string` - so there is no need for a check here.

designs/2022-03-06-representation-of-repeatable-starlark-flags.md
fmeum: We will want to remove `allow_multiple` by introducing and later flipping a flag such as `--incompatible_no_allow_multiple`.

designs/2022-03-06-representation-of-repeatable-starlark-flags.md
fmeum: I added code blocks with examples. The behavior with `repeatable = True` is identical to how `allow_multiple = True` works today, it's only the representation of the config type within Bazel that changes.

designs/2022-03-06-representation-of-repeatable-starlark-flags.md
```diff
@@ -61,28 +61,37 @@ This approach to repeatable flags has the following drawbacks:
...
-  `config.string` flag with `allow_multiple = True`.
+  `config.string` flag with `allow_multiple = True`:
+  
+```
+--copt=a,b,c        --> ["a,b,c"]
+--copt=a --copt=b,c --> ["a", "b,c"]
```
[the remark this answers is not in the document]
fmeum: It could, but I think that it shouldn't: 

If we pass `["a", "b,c"]` to the build setting implementation function unprocessed, it is free to turn this into `["a", "b", "c"]` and pass it on to consumers via a provider in that form.

But if we pass `["a", "b", "c"]` directly, we forcibly remove information that the build setting implementation function wouldn't be able to recover. For example, this would mess up linker flags such as `-Wl,--enable-auto-import`.

Given that build settings can already do any post-processing they require in Starlark, I would opt for doing as little pre-processing as possible within Bazel.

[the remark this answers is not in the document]
aranguyen: Thanks for the explanation. I'm not familiar with the linker flags mentioned. Could you elaborate on how it would mess them up?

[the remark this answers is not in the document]
fmeum: If you have a (hypothetical) Starlark version of the native `--copt`, then you may want to pass in `-Wl,--enable-auto-import` as a single argument. But if all repeated flags also split on `,`, then this would be impossible: The final list of args would always include two separate args `-Wl` and `--enable-auto-import`, which changes semantics (in fact, in the case of this particular flag, this would likely lead to a compiler error).

aranguyen (approved):
This looks good to me.

fmeum:
> This looks good to me.

Thanks! Could you merge? I don't have the rights.

aranguyen:
Done! FYI, I'm working on your PR and will update as soon as I can.

PR #269, opened 2022-07-21
Add "Locating runfiles with Bzlmod"

fmeum:


fmeum:
@Wyverald @meteorcloudy @phst I have added you as reviewers based on the discussion we had about this topic. I would like to have a maintainer of either rules_java or rules_cc on the list as well, but wasn't sure who would be available.

I'm a bit uncertain about the wording in https://bazel.build/contribute/design-documents#new-proposal. Should I already send out the email or wait for this PR to be merged? "Submitted" could mean either of the two.

fmeum:
@comius Looks like you are back, would you be willing to review this proposal from the rules owners perspective?

designs/2022-07-21-locating-runfiles-with-bzlmod.md
```diff
@@ -0,0 +1,249 @@
...
+
+At runtime, an executable target has to be able to translate an apparent repository name to a canonical repository name.
+This requires serializing the repository mapping of every transitive dependency of the target that may perform runfiles lookups to a file that can be accessed at runtime, similar to the existing runfiles manifest.
+
+In more formal terms, the complete repository mapping maintained by Bazel is a function `repo_mapping` that maps a pair `(C, A)` of a canonical repository name `C` and an apparent repository name `A` to the canonical repository name `repo_mapping(C, A)` to which `A` resolves when referred to from within `C`.
+
+The repository mapping manifest would then look as follows:
+
+1. The repository mapping manifest is an ASCII file named `<executable>.repo_mapping` and placed in the same directory as the target's executable.
+2. Every triple `(C, A, repo_mapping(C, A))` is written as a single line terminated by `\n`, with the three components separated by a single space.
```
Wyverald, line 76: should we use commas as separators instead? There's less room for confusion if the empty string ever shows up, and it also conveniently doubles as a CSV file.

designs/2022-07-21-locating-runfiles-with-bzlmod.md
```diff
@@ -0,0 +1,249 @@
...
+   This is necessary since the main repository is stored under this name.
+3. The lines of the manifest are sorted lexicographically.
+   This ordering is equivalent to the lexicographical ordering of the triples comprising the repository mapping.
+4. The repository mapping manifest for an executable target `T` consists only of those entries `(C, A, repo_mapping(C, A))` for which both of the following conditions are satisfied:
+    - `T` transitively depends on a target in `C` that directly depends on a target advertising `RunfilesLibraryInfo`;
+    - `T`'s runfiles contain an artifact owned by a target in `repo_mapping(C, A)`.
+   This property ensures that the repository mapping manifest only contains the entries that are actually needed for the target to resolve its runfiles.
+   If all rather than this limited set of entries were emitted into the manifest, all actions depending on T would have to rerun if any repository containing a transitive dependency of T would declare a new dependency or change an apparent repository name, even if that repository neither contributes a runfile nor looks up runfiles at runtime.
```
Wyverald, line 86: nit: "If all entries (instead of just this limited set) were emitted" or something. "all rather than this limited set of entries" reads a bit weird.

designs/2022-07-21-locating-runfiles-with-bzlmod.md
```diff
@@ -0,0 +1,249 @@
...
+
+In order to help runfiles libraries find the repository mapping manifest, it is added to the runfiles manifest (and thus the runfiles directory) under the fixed path `_repo_mapping`.
```
Wyverald, line 88: does this mean we emit this file twice, once next to the executable (as `<executable>.repo_mapping`), and once in the runfiles directory? Might be worth calling out if so.

designs/2022-07-21-locating-runfiles-with-bzlmod.md
```diff
@@ -0,0 +1,249 @@
...
+Since user-supplied repository names cannot start with `_`, there is no risk of collision with actual runfiles.
```
Wyverald, line 89: There seems to be an assumption here that the runfiles directory contains directory named after apparent repo names (`A` in the terminology above). Is that true? I was under the impression that they were named after canonical repo names (`repo_mapping(C, A)`), but I don't know runfiles very well...

designs/2022-07-21-locating-runfiles-with-bzlmod.md
```diff
@@ -0,0 +1,249 @@
...
+       "@my_protobuf//:protoc",
+   ],
+   deps = ["@bazel_tools//tools/cpp/runfiles"],
+)
+```
+
+With `my_module` as the root module and Bzlmod enabled, the repository mapping manifest of `//:some_tool` would look as follows (spaces are represented as `[ ]` for clarity):
+
+```
+[ ]my_module[ ]@my_module
```
Wyverald, line 130: per line 78, this should read `[ ]my_module[ ]@my_workspace`, no? Same for the next line

designs/2022-07-21-locating-runfiles-with-bzlmod.md
```diff
@@ -0,0 +1,249 @@
...
+```
+[ ]my_protobuf[ ]@protobuf~3.19.2
+```
+
+This manifest does *not* contain any entries where `repo_mapping(C, A)` is `my_workspace` since the target does not include any runfiles from the main repository.
+
+If a module `other_module` depends on `my_module` and contains a target that depends on `@my_module//:some_tool`, then that target's repository mapping manifest would contain the following lines (among others):
+
+```
+@my_module~1.2.3[ ]my_module[ ]@my_module~1.2.3
```
Wyverald, line 149: just as a heads-up -- as part of my effort to fix https://github.com/bazelbuild/bazel/issues/15916, I might be able to remove the extra '@' that {we prepend to the canonical repo names when Bzlmod is enabled}, and just add the '@' when stringifying a Label instead. Hopefully that shouldn't be disruptive to your work here.

(The current "workaround" of having an extra '@' in the canonical repo names when Bzlmod is enabled is sometimes a bit unsound -- for example, you can't write `@@//foo:bar` to refer to the main repo)

Wyverald (commented):
Thanks again for the great writeup :) I'm still reviewing; leaving first round of comments

designs/2022-07-21-locating-runfiles-with-bzlmod.md
fmeum: This I wasn't sure about: What is the general algorithm to determine the canonical repository name of the main repository? Is it always the workspace name or can it also be the module name? Line 78 is intended to match this.

designs/2022-07-21-locating-runfiles-with-bzlmod.md
fmeum: In the implementation, I would use `RepositoryName#getName()` and hope that that would always return the correct name. But I'll take another look at what the runfiles logic uses.

designs/2022-07-21-locating-runfiles-with-bzlmod.md
```diff
@@ -0,0 +1,249 @@
...
+    - `T` transitively depends on a target in `C` that directly depends on a target advertising `RunfilesLibraryInfo`;
+    - `T`'s runfiles contain an artifact owned by a target in `repo_mapping(C, A)`.
+   This property ensures that the repository mapping manifest only contains the entries that are actually needed for the target to resolve its runfiles.
+   If all entries (instead of just this limited set) were emitted into the manifest, all actions depending on T would have to rerun if any repository containing a transitive dependency of T would declare a new dependency or change an apparent repository name, even if that repository neither contributes a runfile nor looks up runfiles at runtime.
+
+In order to help runfiles libraries find the repository mapping manifest, it is added to the runfiles manifest (and thus the runfiles directory) under the fixed path `_repo_mapping`.
+Since canonical repository names do not start with `_`, there is no risk of this synthetic part being a prefix of an actual runfile.
```
Wyverald, line 89: this part is potentially untrue -- canonical repo names _can_ start with `_`. We currently don't have such repos (not that I know of, anyway), but we haven't ruled out the possibility. (For example, there's a pseudo-repo called `@_builtins`.)

Is the runfiles manifest also inside the runfiles directory, or is it only outside of it? Is it crucial for the `_repo_mapping` file to be inside the runfiles directory? Maybe I'm trying too hard to be 100% sure that a conflict can't happen...

designs/2022-07-21-locating-runfiles-with-bzlmod.md
Wyverald: The canonical repository name of the main repository is always "". Both "my_module" and "my_workspace" should be aliases for it (i.e. map to "").

A different concept (the "workspace name") is used for exec file paths (and presumably the runfiles directory) too. That one is always the name specified in WORKSPACE (as of right now anyway), or if not specified, `__main__`.

designs/2022-07-21-locating-runfiles-with-bzlmod.md
Wyverald: yeah, that should be fine. As long as there's no built-in assumption that the canonical repo name starts with an '@' or something.

aiuto:
> I'm a bit uncertain about the wording in https://bazel.build/contribute/design-documents#new-proposal. Should I already send out the email or wait for this PR to be merged? "Submitted" could mean either of the two.

Since we seem to be reviewing the proposal in code review rather than a doc, you should send mail to bazel-dev@ now, so others can see it without stalking this repo. I always avoid the problem by doing the proposal in a doc and making the PR just be a pointer to the doc. I find github's UI really insufficient for having discussions about text.

designs/2022-07-21-locating-runfiles-with-bzlmod.md
```diff
@@ -0,0 +1,249 @@
...
+
+Furthermore, depending on the nature of the programming language of the target, the language-specific rules can use this information to generate code that allows the target compile-time access to its canonical repository name as described in Change 4.
+
+### Backwards compatibility
+
+Adding a new global symbol is a backwards-compatible change.
+
+In order to support Bzlmod, rulesets providing runfiles libraries will have to be updated to advertise the new provider.
+Since a new global symbol can't be feature-detected in Starlark, this can only be done once the ruleset providing the runfiles libraries targets a Bazel version that offers the new provider as its minimum supported version.
+For runfiles libraries that do not ship with Bazel, it would thus be helpful to cherry-pick this change into Bazel 5.
```
aiuto, line 64: Overall, it would be helpful if all runfiles libraries were external to Bazel.  That facilitates writing code that gets runfiles under bazel test and bazel run, but can also be repackaged and installed in the canonical manner for any OS. Those rarely would have a .runfiles tre e the way we build it.

Even if we make them all external, we may not have to backport this feature to bazel 5.x.   Since each Bazel LTS could have the potential to do very different things with runfiles, we will have to account for that at the library side. Either we have versions of library for each LTS track, or, have the runfiles manifest/layout itself contain a version number.  Some day in the future, Bazel 9.x will use runfiles-v2, and a single runfiles library at that time could read all back versions.  That style might solve the problem of backport to 5.x  We could use the non-existance of the runfiles version to know it is the old style, and read accordingly.

fmeum: I understand the appeal of having runfiles be external to Bazel, but in this particular section, a backport of a small part of the functionality proposed here seems actually needed even if all libraries became external already tomorrow: We have to provide a way for runfiles libraries to mark themselves as such and this way has to be understood by Bazel itself.

That's because we are essentially introducing version 1 of a new manifest (the repository mapping manifest) here rather than a new version of an existing manifest. The new manifest type needs some additional support within Bazel itself.

At my company, we are actually using the runfiles libraries as they are provided right now at install time by relying on the manifest- rather than the directory-based implementation.

designs/2022-07-21-locating-runfiles-with-bzlmod.md
fmeum: On Windows without symlink support, the runfiles directory usually isn't created (it would be very expensive). Instead, the manifest is used to look up paths instead.

designs/2022-07-21-locating-runfiles-with-bzlmod.md
fmeum: The important bit for me was that we can't have collisions with canonical repository names of *user-provided repos*. 

Collisions with internally defined repositories are less of a concern since we know them all. Also, nothing is stopping us from giving essentially arbitrary names to Bazel-defined repositories. But if you prefer a different escaping scheme, I can use `@@repo_mapping` or even `_never_let_this_internal_repo_name_through_code_review_repo_mapping` instead. ;-)

designs/2022-07-21-locating-runfiles-with-bzlmod.md
fmeum: Thanks, I clarified this part.

fmeum:
> Since we seem to be reviewing the proposal in code review rather than a doc, you should send mail to bazel-dev@ now, so others can see it without stalking this repo. I always avoid the problem by doing the proposal in a doc and making the PR just be a pointer to the doc. I find github's UI really insufficient for having discussions about text.

Thanks for the clarification, will send out the mail right away.

I also find comments on Google Docs much more convenient than on GitHub PRs, but I prefer Markdown (especially its code blocks) to Docs formatting. It's also nice that Markdown files checked into the repo can't get lost - regaining edit rights to the original runfiles libraries design doc took quite some effort.

Wyverald (approved):

designs/2022-07-21-locating-runfiles-with-bzlmod.md
```diff
@@ -0,0 +1,249 @@
...
+Based on Change 2, runfiles libraries can now apply the repository mapping at runtime.
+Concretely, runfiles libraries should modify their current lookup procedure described in the [original design doc](https://docs.google.com/document/d/e/2PACX-1vSDIrFnFvEYhKsCMdGdD40wZRBX3m3aZ5HhVj4CtHPmiXKDCxioTUbYsDydjKtFDAzER5eg7OjJWs3V/pub) as follows:
+
+1. After the `Create` method of the runfiles library has located either the runfiles directory or the runfiles manifest, it uses either to look up the repository mapping manifest under the fixed runfiles path `_repo_mapping`.
+
+   a. If the runfiles library is shipped with Bazel, it should fail with a descriptive error message if the manifest is not found.
+
+   b. If the runfiles library is not shipped with Bazel, it should fall back to the original lookup procedure. This is necessary to support the use of runfiles libraries with versions of Bazel that do not yet create the manifest.
+
+2. The runfiles library parses the repository mapping manifest and stores a map of `A` to `repo_mapping(C, A)` for the fixed canonical repository name `C` corresponding to the repository from which the `Create` method was called.
```
phst, line 177: Maybe it would be better to determine C in Rlocation, not Create? Otherwise, a runfiles object passed across repository boundaries or a single global lazily-initialized runfiles object would give wrong results when used from different repositories.

fmeum: You are probably correct that this is what we should do for consistency (also with languages like `Bash` that don't even have a `Create`).

The one drawback of that approach is that it makes it more difficult to migrate to the new API as the potentially many `Rlocation` calls all have to be updated. But if that turns out to be a concern, we could always add a `CreateForSingleRepository` function that locks in the repo mapping.

I will change the section.

fmeum: Changed the section, PTAL.

designs/2022-07-21-locating-runfiles-with-bzlmod.md
```diff
@@ -0,0 +1,245 @@
...
+       "@my_protobuf//:protoc",
+   ],
+   deps = ["@bazel_tools//tools/cpp/runfiles"],
+)
+```
+
+With `my_module` as the root module and Bzlmod enabled, the repository mapping manifest of `//:some_tool` would look as follows:
+
+```
+,my_module,my_workspace
```
oquenchil, line 130: Can you clarify that the following:
,my_module,my_workspace
,my_workspace,my_workspace

entries are needed because of data.txt in the data attribute. It took me some time to figure that out.

designs/2022-07-21-locating-runfiles-with-bzlmod.md
```diff
@@ -0,0 +1,245 @@
...
+   This is necessary since the main repository is stored under this name.
+3. The lines of the manifest are sorted lexicographically.
+   This ordering is equivalent to the lexicographical ordering of the triples comprising the repository mapping.
+4. The repository mapping manifest for an executable target `T` consists only of those entries `(C, A, repo_mapping(C, A))` for which both of the following conditions are satisfied:
+    - `T` transitively depends on a target in `C` that directly depends on a target advertising `RunfilesLibraryInfo`;
+    - `T`'s runfiles contain an artifact owned by a target in `repo_mapping(C, A)`.
+   This property ensures that the repository mapping manifest only contains the entries that are actually needed for the target to resolve its runfiles.
+   If all entries (instead of just this limited set) were emitted into the manifest, all actions depending on T would have to rerun if any repository containing a transitive dependency of T would declare a new dependency or change an apparent repository name, even if that repository neither contributes a runfile nor looks up runfiles at runtime.
```
oquenchil, line 86: All of this seems very well thought out but I myself am having trouble imagining how all of this will fit together and foreseeing problems like the one mentioned here without actually looking at the final code. Would it be possible to have a prototype of the whole thing working for at least 2 main rules like Java and C++? I think that would make it easier to spot any issues in the design that cause big problems like unnecessary rerunning actions.

designs/2022-07-21-locating-runfiles-with-bzlmod.md
```diff
@@ -0,0 +1,245 @@
...
+If a module `other_module` depends on `my_module` and contains a target that depends on `@my_module//:some_tool`, then that target's repository mapping manifest would contain the following lines (among others):
+
+```
+@my_module~1.2.3,my_module,@my_module~1.2.3
+@my_module~1.2.3,my_protobuf,@protobuf~3.19.2
+```
+
+### Implementation details
+
+- Add a new internal `RunfilesLibraryUsersProvider` to track the `RepositoryName` and `RepositoryMapping` of transitive dependencies of a given target that directly depend on a runfiles library.
```
oquenchil, line 155: It feels like you have a good reason for this to be its own separate provider and not reuse the existing Runfiles providers to put the mappings there but I don't grasp it completely. Can you call that out and state explicitly the reasons for a separate provider?

designs/2022-07-21-locating-runfiles-with-bzlmod.md
```diff
@@ -0,0 +1,245 @@
...
+
+This fallback behavior is not prone to ambiguity: With Bzlmod, canonical names of non-main repositories always starts with an `@`, which is never contained in a valid apparent repository name.
+
+### Backwards compatibility
+
+End users that do not use Bzlmod will experience no changes when using updated runfiles libraries.
+
+End users that do use Bzlmod currently can't use runfiles libraries successfully in all but the simplest cases (lookups of runfiles contained in the main repository), thus backwards compatibility is not a concern.
+
+# Suggested follow-up changes
```
oquenchil, line 208: Are we leaving anything behind that gets deprecated and would require house-keeping? Assuming a world where everyone is using Bzlmod, is anything left in the code base that is no longer necessary that would need to be cleaned up? If that's the case, can you have a section describing it?

designs/2022-07-21-locating-runfiles-with-bzlmod.md
fmeum: It should be possible to just add a field to `RunfilesProvider`. In the end I expect this to come down to what causes less retained heap usage - I simply don't know yet. 

I added this to the section as an open question.

designs/2022-07-21-locating-runfiles-with-bzlmod.md
fmeum: I can't think of anything. Even though Bzlmod is the motivation for this proposal, there is no direct dependency between the two as repository mappings can already be used in `WORKSPACE` and would break Stardoc today. It just so happens that they are very rarely used (although I believe rules_kotlin relies on them).

designs/2022-07-21-locating-runfiles-with-bzlmod.md
fmeum: I have a prototype for this for Java. I will look into developing it a bit more and including C++.

designs/2022-07-21-locating-runfiles-with-bzlmod.md
fmeum: Thought about it some more and updated the section: The `RunfilesProvider` is controlled by rules implementations, but this information has to be forwarded in all cases, regardless of whether the rules cooperate or not.

fmeum:
@oquenchil and everybody else who is interested in a demo: I created an [example repo](https://github.com/fmeum/bzlmod-runfiles-example) that you can try out the main parts of this proposal on using the linked [rough prototype of a Bazel implementation](https://github.com/fmeum/bazel/tree/bzlmod-runfiles) of parts 1, 2, and 4 for Java and C++. The example binaries emit their repository mapping manifests as well as the value of the injected current repository name.

designs/2022-07-21-locating-runfiles-with-bzlmod.md
oquenchil: By non-cooperating rules do I understand correctly that these are rules that do not propagate downstream the runfiles of its dependencies or place anything in runfiles themselves. If they do not have runfiles, then why is the mapping necessary?

fmeum: If the rule doesn't expose any runfiles, then you are right that the information wouldn't be needed.

I'm worried about the case where the rule forwards runfiles from dependencies without creating a new runfiles object: In that case, there wouldn't be a chance to add its repo mapping if it uses runfiles libraries. That case is why I'm leaning towards a separate provider.

oquenchil: If it forwards the runfiles from the dependencies, as far as I understand it has to merge those dependencies. Couldn't this be part of the merge() method that creates the new RunfilesProvider?

fmeum: This is true for all sensible rule designs I can think of, but it's not strictly required: A rule that e.g. only allows a single dependency could conceivably get by without using the `merge` method in its implementation. But of course that would be pretty pathological and break only if that dependency happens to be a runfiles library.

I'm more and more inclined to switch to just using a field on `RunfilesProvider` and accept that targets based on pathological rules may no be able to access runfiles. We can probably solve this with documentation, what do you think?

oquenchil: Sorry for the silence the past few days. Yes, I agree let's do a field in RunfilesProvider.

oquenchil:
LGTM

fmeum:
@oquenchil Thanks for the review, I changed the relevant section to state that we are planning to reuse `RunfilesProvider`. I also listed you as a reviewer.

Do you happen to see a way we could introduce `RunfilesLibraryInfo` as a Starlark-defined provider? I tried this in https://github.com/bazelbuild/bazel/pull/16077#issuecomment-1210665666, but hit the problem that builtins can't access anything under `@bazel_tools`.

designs/2022-07-21-locating-runfiles-with-bzlmod.md
```diff
@@ -0,0 +1,235 @@
...
+   This is a problem since the path contains the version of `bar`, which isn't static, and thus can't be referenced as a constant literal string anymore.
+   In fact, unless `foo` is the main module, the same problem already arises for files in `foo` itself.
+   This issue is tracked by [bazelbuild/bazel#15029](https://github.com/bazelbuild/bazel/issues/15029).
+2. [Stardoc](https://github.com/bazelbuild/stardoc) generates documentation from `.bzl` files and for that has to resolve labels in `load` statements to `.bzl` files on disk.
+   Since it doesn't actually evaluate `WORKSPACE` or `MODULE.bazel` files, it has no way of discovering repository mappings in the same way as Bazel.
+   Thus, with Bzlmod, Stardoc fails to generate documentation for any `.bzl` file that loads `.bzl` files from other repositories.
+   This issue is tracked by [bazelbuild/bazel#14140](https://github.com/bazelbuild/bazel/issues/14140) and [bazelbuild/stardoc#117](https://github.com/bazelbuild/stardoc/issues/117).
```
tetromino, line 45: FYI: my current plan (currently in early prototype) is for stardoc to avoid doing any kind of custom disk paths munging, and instead, read documentation from the compiled starlark module object stored in skyframe through exactly the same mechanism as what is currently used for implementing the `load` statement.

In other words: if you can `load` the .bzl file, you'll be able to generate docs from it.

fmeum: That sounds great! Could you keep me updated on this on https://github.com/bazelbuild/bazel/issues/16124?

PR #275, opened 2022-09-11
Cross-platform native launchers for Python

groodt:
This document describes an approach for launching `py_binary` artifacts hermetically using the resolved python toolchain.

cc @meteorcloudy @rickeylev @jesseschalken

### DIFF
--- a/designs/2022-09-12-native-launchers-python.md
+++ b/designs/2022-09-12-native-launchers-python.md
@@ -20,7 +20,7 @@ This document describes an approach for launching `py_binary` artifacts hermetic
 Currently, `py_binary` is non-hermetic and launches inconsistently between platforms.
 
 On macos and Linux, there is a [python_stub](https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/bazel/rules/python/python_stub_template.txt)
-that is non-hermetic and requires a "bootstrap" Python interpreter on the host. The "shebang" can be overridden, but
+that is non-hermetic and requires a bootstrap Python interpreter on the host. The "shebang" can be overridden, but
 a "shebang" is always dependent on the runtime host.
 
 On Windows, there is a [native launcher](https://github.com/meteorcloudy/bazel/blob/master/src/tools/launcher/python_launcher.cc)
--- a/designs/2022-09-12-native-launchers-python.md
+++ b/designs/2022-09-12-native-launchers-python.md
@@ -24,7 +24,7 @@ that is non-hermetic and requires a "bootstrap" python interpreter on the host. 
 a "shebang" is always dependent on the runtime host.
 
 On Windows, there is a [native launcher](https://github.com/meteorcloudy/bazel/blob/master/src/tools/launcher/python_launcher.cc)
-that launches `python.exe` on the host which then launches the `py_binary` with the same `python_stub` as macos and linux.
+that launches `python.exe` on the host which then launches the `py_binary` with the same `python_stub` as macos and Linux.
 
 Related issues:
 * [py_binary with hermetic toolchain requires a system interpreter](https://github.com/bazelbuild/rules_python/issues/691)
--- a/designs/2022-09-12-native-launchers-python.md
+++ b/designs/2022-09-12-native-launchers-python.md
@@ -32,7 +32,7 @@ Related issues:
 
 This situation is undesirable because it assumes that the target platform has a bootstrapping python interpreter 
 available and makes the hermetic Python interpreters available with `rules_python` less useful. It is also surprising to 
-users who expect bazel to output self-contained binary artifacts for a target platform.
+users who expect Bazel to output self-contained binary artifacts for a target platform.
 
 The reason this situation exists is because of "bootstrapping". Ultimately, *something* needs to find the python
 interpreter in the runfiles and use that to launch the program. Currently, Bazel assumes the target platform will
--- a/designs/2022-09-12-native-launchers-python.md
+++ b/designs/2022-09-12-native-launchers-python.md
@@ -42,7 +42,7 @@ be able to provide the bootstrapping functionality.
 # Proposal
 
 Extend the native launcher functionality to all platforms and use it to locate the relevant Python interpreter and 
-python program in the `runfiles` tree to launch the `py_binary`. No assumptions should be made about the target platform.
+Python program in the `runfiles` tree to launch the `py_binary`. No assumptions should be made about the target platform.
 
 In pseudo-code, the proposal is as follows:
 
--- a/designs/2022-09-12-native-launchers-python.md
+++ b/designs/2022-09-12-native-launchers-python.md
@@ -69,7 +69,7 @@ This proposal would aim to go further and have these launchers available on all 
 Once this proposal is implemented, it would enable cross-builds of hermetic `py_binary` for all major platforms. It
 would also remove the complexity introduced by having so many chains of nested execution to launch a Python program.
 
-Finally, while this proposal is specific to python, this solution could perhaps be reused for `java_binary`, `sh_binary`
+Finally, while this proposal is specific to Python, this solution could perhaps be reused for `java_binary`, `sh_binary`
 and perhaps be made available for any custom rules that require an interpreter to launch.
 
 
--- a/designs/2022-09-12-native-launchers-python.md
+++ b/designs/2022-09-12-native-launchers-python.md
@@ -1,0 +1,79 @@
+---
+created: 2022-09-12
+last updated: 2022-09-12
+status: To be reviewed
+reviewers:
+  - TODO
+title: Cross-platform native launchers for Python
+authors:
+  - groodt
+---
+
+
+# Abstract
+
+This document describes an approach for launching `py_binary` artifacts hermetically using the resolved python toolchain.
+
+
+# Background
+
+Currently, `py_binary` is non-hermetic and launches inconsistently between platforms.
+
+On macos and Linux, there is a [python_stub](https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/bazel/rules/python/python_stub_template.txt)
+that is non-hermetic and requires a "bootstrap" python interpreter on the host. The "shebang" can be overridden, but
+a "shebang" is always dependent on the runtime host.
+
+On Windows, there is a [native launcher](https://github.com/meteorcloudy/bazel/blob/master/src/tools/launcher/python_launcher.cc)
+that launches `python.exe` on the host which then launches the `py_binary` with the same `python_stub` as macos and linux.
+
+Related issues:
+* [py_binary with hermetic toolchain requires a system interpreter](https://github.com/bazelbuild/rules_python/issues/691)
+* [Neither python_top nor python toolchain works with starlark actions on windows](https://github.com/bazelbuild/bazel/issues/7947#issuecomment-495265016)
+
+This situation is undesirable because it assumes that the target platform has a bootstrapping python interpreter 
+available and makes the hermetic python interpreters available with `rules_python` less useful. It is also surprising to 
+users who expect bazel to output self-contained binary artifacts for a target platform.
+
+The reason this situation exists is because of "bootstrapping". Ultimately, *something* needs to find the python
+interpreter in the runfiles and use that to launch the program. Currently, bazel assumes the target platform will
+be able to provide the "bootstrapping" functionality.
+
+
+# Proposal
+
+Extend the native launcher functionality to all platforms and use it to locate the relevant python interpreter and 
+python program in the `runfiles` tree to launch the `py_binary`. No assumptions should be made about the target platform.
+
+In pseudo-code, the proposal is as follows:
+
+```
+exec(env, runfiles-interpreter, ["interpreter_arg1",], "main.py", ["arg1",])
+```
+
+| Token                  | Description |
+| ---------------------- | ----------- |
+| env                    | Dictionary of key-value pairs for the environment of the process       |
+| runfiles-interpreter   | The resolved python toolchain in runfiles        |
+| ["interpreter_arg1",]  | An array of arguments to provide to the python interpreter        |
+| "main.py"              | The python program to launch in runfiles        |
+| ["arg1",]              | An array of arguments to provide to the python program as sys.argv[1:]        |
+
+This native launcher idea has been proposed a few times by bazel contributors and the community:
+* [Greg Roodt (Community)](https://github.com/bazelbuild/rules_python/issues/691#issuecomment-1174935972)
+* [Yun Peng (Google)](https://github.com/bazelbuild/bazel/issues/7947#issuecomment-495265016)
+* [Richard Levasseur (Google)](https://github.com/bazelbuild/rules_python/issues/691#issuecomment-1186379617)
+
+Some related work has been done that fixes Linux to Windows cross-builds of the Windows launcher. See: [Fix Linux to Windows cross compilation of py_binary, java_binary and sh_binary using MinGW](https://github.com/bazelbuild/bazel/pull/16019)
+This proposal would aim to go further and have these launchers available on all platforms to that a Windows to Linux 
+launcher build could occur (assuming appropriate toolchains).
+
+Once this proposal is implemented, it would enable cross-builds of hermetic `py_binary` for all major platforms. It
+would also remove the complexity introduced by having so many chains of nested execution to launch a python program.
+
+Finally, while this proposal is specific to python, this solution could perhaps be reused for `java_binary`, `sh_binary`
+and perhaps be made available for any custom rules that require an interpreter to launch.
+
+
+# Backward-compatibility
+
+This proposal could require users to setup a cc toolchain for remote execution.
--- a/designs/2022-09-12-native-launchers-python.md
+++ b/designs/2022-09-12-native-launchers-python.md
@@ -31,7 +31,7 @@ Related issues:
 * [Neither python_top nor python toolchain works with starlark actions on windows](https://github.com/bazelbuild/bazel/issues/7947#issuecomment-495265016)
 
 This situation is undesirable because it assumes that the target platform has a bootstrapping python interpreter 
-available and makes the hermetic python interpreters available with `rules_python` less useful. It is also surprising to 
+available and makes the hermetic Python interpreters available with `rules_python` less useful. It is also surprising to 
 users who expect bazel to output self-contained binary artifacts for a target platform.
 
 The reason this situation exists is because of "bootstrapping". Ultimately, *something* needs to find the python
--- a/designs/2022-09-12-native-launchers-python.md
+++ b/designs/2022-09-12-native-launchers-python.md
@@ -67,7 +67,7 @@ Some related work has been done that fixes Linux to Windows cross-builds of the 
 This proposal would aim to go further and have these launchers available on all platforms, including cross_builds where appropriate toolchains are in place.
 
 Once this proposal is implemented, it would enable cross-builds of hermetic `py_binary` for all major platforms. It
-would also remove the complexity introduced by having so many chains of nested execution to launch a python program.
+would also remove the complexity introduced by having so many chains of nested execution to launch a Python program.
 
 Finally, while this proposal is specific to python, this solution could perhaps be reused for `java_binary`, `sh_binary`
 and perhaps be made available for any custom rules that require an interpreter to launch.
--- a/designs/2022-09-12-native-launchers-python.md
+++ b/designs/2022-09-12-native-launchers-python.md
@@ -34,7 +34,7 @@ This situation is undesirable because it assumes that the target platform has a 
 available and makes the hermetic Python interpreters available with `rules_python` less useful. It is also surprising to 
 users who expect Bazel to output self-contained binary artifacts for a target platform.
 
-The reason this situation exists is because of "bootstrapping". Ultimately, *something* needs to find the Python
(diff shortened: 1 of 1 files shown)


meteorcloudy:
@rickeylev Can you review this proposal when you have time ;)

designs/2022-09-12-native-launchers-python.md
```diff
@@ -0,0 +1,78 @@
...
+
+This document describes an approach for launching `py_binary` artifacts hermetically using the resolved python toolchain.
+
+
+# Background
+
+Currently, `py_binary` is non-hermetic and launches inconsistently between platforms.
+
+On macos and Linux, there is a [python_stub](https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/bazel/rules/python/python_stub_template.txt)
+that is non-hermetic and requires a "bootstrap" python interpreter on the host. The "shebang" can be overridden, but
```
rickeylev, line 23: re: "requires a "bootstrap" python interpreter on the host": Well, not necessarily.

It's possible to make the stub use the same runtime as the program it ultimately invokes. This is a pretty sensible thing to do, I think. This is fine if you're already using a system interpreter (you already have an absolute path)

The problem case are "in build" interpreters -- you have to put a relative path to the interpreter. This makes running the program sensitive to the CWD you start the program in. Transforming that relative path into an absolute path requires some sort of runtime logic.

designs/2022-09-12-native-launchers-python.md
```diff
@@ -0,0 +1,78 @@
...
+Related issues:
+* [py_binary with hermetic toolchain requires a system interpreter](https://github.com/bazelbuild/rules_python/issues/691)
+* [Neither python_top nor python toolchain works with starlark actions on windows](https://github.com/bazelbuild/bazel/issues/7947#issuecomment-495265016)
+
+This situation is undesirable because it assumes that the target platform has a bootstrapping python interpreter 
+available and makes the hermetic python interpreters available with `rules_python` less useful. It is also surprising to 
+users who expect bazel to output self-contained binary artifacts for a target platform.
+
+The reason this situation exists is because of "bootstrapping". Ultimately, *something* needs to find the python
+interpreter in the runfiles and use that to launch the program. Currently, bazel assumes the target platform will
```
rickeylev, line 38:
```suggestion
interpreter in the runfiles and use that to launch the program. Currently, Bazel assumes the target platform will
```

designs/2022-09-12-native-launchers-python.md
```diff
@@ -0,0 +1,78 @@
...
+
+On Windows, there is a [native launcher](https://github.com/meteorcloudy/bazel/blob/master/src/tools/launcher/python_launcher.cc)
+that launches `python.exe` on the host which then launches the `py_binary` with the same `python_stub` as macos and linux.
+
+Related issues:
+* [py_binary with hermetic toolchain requires a system interpreter](https://github.com/bazelbuild/rules_python/issues/691)
+* [Neither python_top nor python toolchain works with starlark actions on windows](https://github.com/bazelbuild/bazel/issues/7947#issuecomment-495265016)
+
+This situation is undesirable because it assumes that the target platform has a bootstrapping python interpreter 
+available and makes the hermetic python interpreters available with `rules_python` less useful. It is also surprising to
```
rickeylev, line 34:
```suggestion
available and makes the hermetic Python interpreters available with `rules_python` less useful. It is also surprising to 
```

designs/2022-09-12-native-launchers-python.md
```diff
@@ -0,0 +1,78 @@
...
+# Background
+
+Currently, `py_binary` is non-hermetic and launches inconsistently between platforms.
+
+On macos and Linux, there is a [python_stub](https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/bazel/rules/python/python_stub_template.txt)
+that is non-hermetic and requires a "bootstrap" python interpreter on the host. The "shebang" can be overridden, but
+a "shebang" is always dependent on the runtime host.
+
+On Windows, there is a [native launcher](https://github.com/meteorcloudy/bazel/blob/master/src/tools/launcher/python_launcher.cc)
+that launches `python.exe` on the host which then launches the `py_binary` with the same `python_stub` as macos and linux.
```
rickeylev, line 27:
```suggestion
that launches `python.exe` on the host which then launches the `py_binary` with the same `python_stub` as macos and Linux.
```

designs/2022-09-12-native-launchers-python.md
rickeylev, line 23:
```suggestion
that is non-hermetic and requires a "bootstrap" Python interpreter on the host. The "shebang" can be overridden, but
```

designs/2022-09-12-native-launchers-python.md
```diff
@@ -0,0 +1,78 @@
...
+  - TODO
+title: Cross-platform native launchers for Python
+authors:
+  - groodt
+---
+
+
+# Abstract
+
+This document describes an approach for launching `py_binary` artifacts hermetically using the resolved python toolchain.
```
rickeylev, line 15:
```suggestion
This document describes an approach for launching `py_binary` artifacts hermetically using the resolved Python toolchain.
```

designs/2022-09-12-native-launchers-python.md
```diff
@@ -0,0 +1,78 @@
...
+On Windows, there is a [native launcher](https://github.com/meteorcloudy/bazel/blob/master/src/tools/launcher/python_launcher.cc)
+that launches `python.exe` on the host which then launches the `py_binary` with the same `python_stub` as macos and linux.
+
+Related issues:
+* [py_binary with hermetic toolchain requires a system interpreter](https://github.com/bazelbuild/rules_python/issues/691)
+* [Neither python_top nor python toolchain works with starlark actions on windows](https://github.com/bazelbuild/bazel/issues/7947#issuecomment-495265016)
+
+This situation is undesirable because it assumes that the target platform has a bootstrapping python interpreter 
+available and makes the hermetic python interpreters available with `rules_python` less useful. It is also surprising to 
+users who expect bazel to output self-contained binary artifacts for a target platform.
```
rickeylev, line 35:
```suggestion
users who expect Bazel to output self-contained binary artifacts for a target platform.
```

designs/2022-09-12-native-launchers-python.md
```diff
@@ -0,0 +1,78 @@
...
+
+The reason this situation exists is because of "bootstrapping". Ultimately, *something* needs to find the python
```
rickeylev, line 37:
```suggestion
The reason this situation exists is because of "bootstrapping". Ultimately, *something* needs to find the Python
```

rickeylev, line 37:
```suggestion
The reason this situation exists is because of bootstrapping. Ultimately, *something* needs to find the python
```

No need to use quotes. Quotes like this imply the word has some alternative meaning than it's regular meaning.

designs/2022-09-12-native-launchers-python.md
```diff
@@ -0,0 +1,78 @@
...
+interpreter in the runfiles and use that to launch the program. Currently, bazel assumes the target platform will
+be able to provide the "bootstrapping" functionality.
```
rickeylev, line 39:
```suggestion
be able to provide the bootstrapping functionality.
```

designs/2022-09-12-native-launchers-python.md
```diff
@@ -0,0 +1,78 @@
...
+
+
+# Proposal
+
+Extend the native launcher functionality to all platforms and use it to locate the relevant python interpreter and
```
rickeylev, line 44:
```suggestion
Extend the native launcher functionality to all platforms and use it to locate the relevant Python interpreter and 
```

designs/2022-09-12-native-launchers-python.md
```diff
@@ -0,0 +1,78 @@
...
+
+The reason this situation exists is because of "bootstrapping". Ultimately, *something* needs to find the python
+interpreter in the runfiles and use that to launch the program. Currently, bazel assumes the target platform will
+be able to provide the "bootstrapping" functionality.
+
+
+# Proposal
+
+Extend the native launcher functionality to all platforms and use it to locate the relevant python interpreter and 
+python program in the `runfiles` tree to launch the `py_binary`. No assumptions should be made about the target platform.
```
rickeylev, line 45:
```suggestion
Python program in the `runfiles` tree to launch the `py_binary`. No assumptions should be made about the target platform.
```

designs/2022-09-12-native-launchers-python.md
```diff
@@ -0,0 +1,78 @@
...
+
+In pseudo-code, the proposal is as follows:
```
rickeylev, line 47: This is pretty high-level psuedo-code :).

Something a little more concrete would be better. e.g., it has to find the runfiles directory to resolve the relative path names.

designs/2022-09-12-native-launchers-python.md
```diff
@@ -0,0 +1,78 @@
...
+
+```
+exec(env, runfiles-interpreter, ["interpreter_arg1",], "main.py", ["arg1",])
+```
+
+| Token                  | Description |
+| ---------------------- | ----------- |
+| env                    | Dictionary of key-value pairs for the environment of the process       |
```
rickeylev, line 55: I don't see why env is one of the inputs? This basically implies that the launcher process may need to use a modified environment from the actual program -- what's the motivation case for this? Why would it not just inherit the existing environment?

Ah, one case I just thought of: LD_PRELOAD (or equiv). Basically, a binary might require such a setting and we wouldn't want the launcher itself to use that (and by "might" i mean, we have this feature internally at Google)

designs/2022-09-12-native-launchers-python.md
```diff
@@ -0,0 +1,78 @@
...
+| runfiles-interpreter   | The resolved python toolchain in runfiles        |
```
rickeylev, line 56: The description here doesn't quite make sense with what the arg name implies.

The arg name sounds like a string path. The description is "the python toolchain", which is a complex object.

designs/2022-09-12-native-launchers-python.md
```diff
@@ -0,0 +1,78 @@
...
+This native launcher idea has been proposed a few times by bazel contributors and the community:
+* [Greg Roodt (Community)](https://github.com/bazelbuild/rules_python/issues/691#issuecomment-1174935972)
+* [Yun Peng (Google)](https://github.com/bazelbuild/bazel/issues/7947#issuecomment-495265016)
+* [Richard Levasseur (Google)](https://github.com/bazelbuild/rules_python/issues/691#issuecomment-1186379617)
+
+Some related work has been done that fixes Linux to Windows cross-builds of the Windows launcher. See: [Fix Linux to Windows cross compilation of py_binary, java_binary and sh_binary using MinGW](https://github.com/bazelbuild/bazel/pull/16019)
+This proposal would aim to go further and have these launchers available on all platforms, including cross_builds where appropriate toolchains are in place.
+
+Once this proposal is implemented, it would enable cross-builds of hermetic `py_binary` for all major platforms. It
+would also remove the complexity introduced by having so many chains of nested execution to launch a python program.
```
rickeylev, line 70:
```suggestion
would also remove the complexity introduced by having so many chains of nested execution to launch a Python program.
```

designs/2022-09-12-native-launchers-python.md
```diff
@@ -0,0 +1,78 @@
...
+
+Finally, while this proposal is specific to python, this solution could perhaps be reused for `java_binary`, `sh_binary`
```
rickeylev, line 72:
```suggestion
Finally, while this proposal is specific to Python, this solution could perhaps be reused for `java_binary`, `sh_binary`
```

rickeylev, line 72: We should reach out to those owners to see if they have input to add.

designs/2022-09-12-native-launchers-python.md
```diff
@@ -0,0 +1,78 @@
...
+and perhaps be made available for any custom rules that require an interpreter to launch.
+
+
+# Backward-compatibility
+
+This proposal could require users to setup a cc toolchain for remote execution.
```
rickeylev, line 78: Is it _required_ to be CC? e.g., what if someone wants to write a launcher in rust?
Can a pre-built launcher be used? I guess that _might_ be precluded; it depends on how the argv, path to interpretter etc is embedded or passed along. This makes me wonder if supporting code in Bazel itself would work better -- e.g. return a special provider that says "we're doing a launcher thing, use this argv etc when running"

designs/2022-09-12-native-launchers-python.md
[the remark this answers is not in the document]
groodt: Who would be the most likely person to accept this sort of proposal? I'm happy to expand this beyond Python (I mention in my proposal, that it could be expanded). Im just not sure if it's better to start incrementally and pick Python first, or if it's better to start broad.

designs/2022-09-12-native-launchers-python.md
```diff
@@ -0,0 +1,78 @@
...
+
+This document describes an approach for launching `py_binary` artifacts hermetically using the resolved Python toolchain.
+
+
+# Background
+
+Currently, `py_binary` is non-hermetic and launches inconsistently between platforms.
+
+On macos and Linux, there is a [python_stub](https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/bazel/rules/python/python_stub_template.txt)
+that is non-hermetic and requires a "bootstrap" Python interpreter on the host. The "shebang" can be overridden, but
```
groodt, line 23:
```suggestion
that is non-hermetic and requires a bootstrap Python interpreter on the host. The "shebang" can be overridden, but
```

designs/2022-09-12-native-launchers-python.md
```diff
@@ -0,0 +1,78 @@
...
+
+Related issues:
+* [py_binary with hermetic toolchain requires a system interpreter](https://github.com/bazelbuild/rules_python/issues/691)
+* [Neither python_top nor python toolchain works with starlark actions on windows](https://github.com/bazelbuild/bazel/issues/7947#issuecomment-495265016)
+
+This situation is undesirable because it assumes that the target platform has a bootstrapping python interpreter 
+available and makes the hermetic Python interpreters available with `rules_python` less useful. It is also surprising to 
+users who expect Bazel to output self-contained binary artifacts for a target platform.
+
+The reason this situation exists is because of "bootstrapping". Ultimately, *something* needs to find the Python
```
groodt, line 37:
```suggestion
The reason this situation exists is because of bootstrapping. Ultimately, *something* needs to find the Python
```

designs/2022-09-12-native-launchers-python.md
groodt: > This is fine if you're already using a system interpreter (you already have an absolute path)

This is probably the only case where it's fine and it has other issues. It's not portable unless everyone has the same interpreter installed at the same global path.

It also goes against the premise of the proposal, in the sense that it's "non-hermetic". It requires all hosts to have a preinstalled runtime at a particular version, which isn't easy to guarantee.

designs/2022-09-12-native-launchers-python.md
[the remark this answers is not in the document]
rickeylev: Java is the most immediate one that comes to mind. I'm pretty sure they are basically doing the same thing we are doing today (shell script that does setup, an optional launcher for a native binary). I emailed our internal language-rule-owners group.

I'm primarily interested in collecting advice and assessing interest at this point. I like to keep as small a set of Deciders as possible. Not all advice must be obeyed.

designs/2022-09-12-native-launchers-python.md
```diff
@@ -12,37 +12,37 @@ authors:
...
 # Background
 
 Currently, `py_binary` is non-hermetic and launches inconsistently between platforms.
 
 On macos and Linux, there is a [python_stub](https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/bazel/rules/python/python_stub_template.txt)
-that is non-hermetic and requires a "bootstrap" python interpreter on the host. The "shebang" can be overridden, but
+that is non-hermetic and requires a bootstrap Python interpreter on the host. The "shebang" can be overridden, but
 a "shebang" is always dependent on the runtime host.
 
 On Windows, there is a [native launcher](https://github.com/meteorcloudy/bazel/blob/master/src/tools/launcher/python_launcher.cc)
```
rickeylev, line 26: I think you meant to link to bazelbuild here, not meteorcloudy?

https://github.com/bazelbuild/bazel/blob/master/src/tools/launcher/python_launcher.cc

meteorcloudy:
@rickeylev Yes, porting the launcher to Linux and macOS shouldn't be too hard. There is no fundamental Windows specific dependencies.

rickeylev:
Thanks, that's good to hear.

For posterity, the code in the py rules for this is around here:
https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/bazel/rules/python/BazelPythonSemantics.java#L294

rickeylev:
I asked if it is possible to get the launcher code itself out of Bazel's core. If we can, then great. Otherwise, I'd be +1 on implementing it in rules_python and rewiring py_binary to, somehow, use that instead of the one built into bazel.

meteorcloudy:
Moving the //src/tools/launcher to its own repo is going to be very hard, because the launcher also depends on cc_libraries under //src/main/cpp/util.

But I do think it's a good idea to start a new repo (ideally under bazel-contrib) which implements a launcher that could be used by different rules in the community. This is going to be helpful for porting rules to Windows, because launching a binary on Windows is a problem for almost all rules, but the launcher in Bazel only supports Java, Python, and Bash.

rickeylev:
meteorcloudy and I talked a bit.

We both support the idea of having this code outside of Bazel itself so that it isn't coupled to the Bazel releases. Getting to that point...I'm not clear on that part. I'm willing to entertain ideas -- perhaps a new toolchain the Java impl looks up, but is implemented in rules_python in Starlark? Maybe in combination with some helpers we expose on e.g. py_common? Maybe have the rules_python macro pass something in to the rule? Anyways, I'm _very_ willing to entertain ideas that let us decouple from Bazel core.

I'm also very much OK if we put this in rules_python and deal with cross-language collaboration separately (the impression I've gotten from language owners thus far is "yeah that affects us" but not "we want to be part of the solution"). This is an internal detail of the rules, so it shouldn't be too hard to adapt if a collaboration happens.

I'm also fine with modifying this launcher-code in Bazel itself; I don't own it, so I'd like to know who we send PRs to for approval.

@comius

meteorcloudy:
> I'm also fine with modifying this launcher-code in Bazel itself; I don't own it, so I'd like to know who we send PRs to for approval.

I'm the original author of the launcher, so you can send the PRs to me.

designs/2022-09-12-native-launchers-python.md
```diff
@@ -0,0 +1,78 @@
...
+status: To be reviewed
+reviewers:
+  - TODO
+title: Cross-platform native launchers for Python
+authors:
+  - groodt
+---
+
+
+# Abstract
```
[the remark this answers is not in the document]
oga35767-eng: What idea u got 👍

designs/2022-09-12-native-launchers-python.md
```diff
@@ -0,0 +1,78 @@
...
+
+This document describes an approach for launching `py_binary` artifacts hermetically using the resolved python toolchain.
+
+
+# Background
+
+Currently, `py_binary` is non-hermetic and launches inconsistently between platforms.
```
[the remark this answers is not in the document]
oga35767-eng: 👍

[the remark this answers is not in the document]
oga35767-eng: Don't laugh ok , I jus started picking up on all these , there's so much  to pick from I want to use all of it that I jus give myself a headache , but this is what I am trying to do , I have a lot of music it's generated but its important, and it's all type of music from different artists  but it's based on my life,  system picks up on my energy I think yout idea  good 👍

designs/2022-09-12-native-launchers-python.md
oga35767-eng: What do u mean  , it's a good idea it will figure out the best way to move forward with my project the best way nobody has it , I don't want to tell u why I am doing this yet

designs/2022-09-12-native-launchers-python.md
oga35767-eng: I have no clue what you telling me or what to do next , its hard to explain it but I know what you mean about your idea ,

designs/2022-09-12-native-launchers-python.md
oga35767-eng: Look I don't know how to put all this together but I know what I want to do , I have ideas too I jus don't know how to put them together, so wuz up , how can you help me help you you get this going  I got all the platforms going  cuz the jaking my sht , ideas  , is I need to put my stuff in a place that will keep track of who is using it ,

designs/2022-09-12-native-launchers-python.md
oga35767-eng: If I want something it will choose the best for me if I connect to it it will understand and know who I am anything that has to do with codes recognize me

designs/2022-09-12-native-launchers-python.md
oga35767-eng: I have no idea what you are saying there so much stuff I don't know what to put together what I need , I what to use all of it , I don't talk to ai  do it's not easy

oga35767-eng: But u two can see what I need

PR #277, opened 2022-10-17
Dependency adapters for virtual targets

vdye:
Hi! 👋 This proposal is intended to address https://github.com/bazelbuild/bazel/issues/16380, following the design review process outlined [here](https://bazel.build/contribute/design-documents).

The document itself is approximately divided into "why I want to do this" (Overview, Background, Concepts) and "how I'd like to do this" (Goals, Implementation, Backward Compatibility). Especially in the first part, I'm interested in whether there are any faulty or incomplete assumptions about Bazel (and, if so, how they should be corrected). I'm happy to clarify or rework anything people would like to discuss, especially if there's a better way to get the behavior I'm looking for.

Finally, I haven't contributed to Bazel before (in case that wasn't obvious 😅), so please let me know if there's anything I can/should do differently to better fit the development and review processes of the project!


lberki:
Mind adding a link from the "under review" section of the contents page?

lberki (approved):

lberki:
Thanks! Merging.

PR #282, opened 2022-12-07
Move "Exec platform per toolchain" to approved

comius:


comius:
cc @kotlaja

katre (approved):

lberki:
As long as @katre is fine with this, I'm fine with this also. I will continue the discussion on the design doc because I'd like to map the landscape out a bit more, but given that @katre is more knowledgeable in the area, he is the one who should make the call.

PR #303, opened 2023-03-30
Drop "Toolchainifying proto rules."

lberki:
The landscape has changed since then; it still makes sense to use toolchains in proto rules, but it will probably be done somewhat differently when we get to it.


Yannic (approved):

lberki:
Thanks! @comius, mind also taking a look?

comius (approved):

PR #311, opened 2023-04-03
Mark "Remote Output Service..." as dropped.

lberki:
The associated pull request
https://github.com/bazelbuild/bazel/pull/12823 still seems to see some activity, but Ed and Chi are not responsive, so I suppose this is the right thing to do.


coeuvre:
IIRC, we never had an agreement on how Bazel should interact with FUSE and since we are going to enable BwoB by default for 7.0, maybe this can be dropped. But I would like to hear from Ed.

lberki:
Yep, I wouldn't merge this before I hear from @EdSchouten (unless I don't hear from him for a few days, for the show must go on!)

lberki:
At the very least, the involvement of aspect.dev and the activity on [#12823](https://github.com/bazelbuild/bazel/pull/12823) indicates there is significant need in the community for functionality like this, so I don't want to summarily pull the plug.

I have a few questions to determine what should happen:
* @EdSchouten : How does the interaction between the FUSE daemon and Bazel look like? Is it limited to the RPC service described in the design doc, i.e. does it work without Bazel ever calling file system operations on the FUSE file system? How and when does the FUSE file system actually get mounted? (the design doc doesn't quite mention this)
* @coeuvre : How much work would it take to merge #12823 ? 
* If you can't contribute, maybe someone from aspect.dev could?

The reason why I'm asking these is that given the demand and the limitations of BwoB, I'm amenable to merging this, but I'm also worried about potential interactions with the rest of Bazel (e.g. BwoB, tree artifacts, whether this works on Linux/Mac OS/Windows) and the safe thing to do is not to merge it.

brentleyjones:
@alexeagle

f0rmiga:
I'm the original author of the Aspect CLI. I'm actively looking into it. I'll post back as soon as I have an answer.

meisterT:
@EdSchouten and others: one thing I am concerned about is ongoing maintenance. If we add this functionality to Bazel, people will come to us and report bugs or ask for features in relation to it and we currently don't have the bandwidth or use-case to staff it.

What is the level of commitment from your side to help out with maintaining the functionality?

alexeagle:
The "remote output service" in google3 is ObjFS, right? If I understand correctly this issue is about how other large companies need the same thing. I do see clients who have this need, in addition to Ed's employer.

Funding (a.k.a. "commit to maintain") has come up here. It feels to me that the reason we're lacking funding for this project is that the implementations diverged. It's hard to convince Google managers that you should spend time maintaining a thing Google doesn't need, but Google's solution for this problem isn't available to others.
Is it feasible to converge this? 

1. if BwoB really does solve the use cases, could google3 switch to that from ObjFS (at least in theory, I'm not suggesting that migration would make sense)
2. If not, it suggests others need an ObjFS equivalent as well. gRPC service with ObjFS so that the maintenance work you already do would support community needs as well?

meisterT:
> if BwoB really does solve the use cases, could google3 switch to that from ObjFS (at least in theory, I'm not suggesting that migration would make sense)

Some of our developer builds are indeed on BwoB and not using ObjFs.


> I do see clients who have this need, in addition to Ed's employer.

What drives this need mostly?

alexeagle:
> What drives this need mostly?

The scenario we run into is exactly what @EdSchouten wrote:

> Needing to constantly switch between `--remote_download_minimal`, `--remote_download_toplevel`

At the time you run a build, you may not know yet which files from `bazel-out` you want to use. You might have an existing shell script which does a build and then executes a program from `bazel-out`, or you might have a wiki document that tells engineers how to find files in that folder after a build. Such a user is currently forced to download the entire build output, even though the files they need are trivially fetchable from the CAS in the remote cache if you only knew the key used to fetch them.

A specific case of wanting to read arbitrarily from `bazel-out` is mentioned above; in the Aspect CLI we support continuous delivery with an [`outputs`](https://github.com/aspect-build/aspect-cli/blob/main/docs/aspect_outputs.md) command which supports a pseudo-mnemonic (`ExecutableHash`) to compute a hash of a binary and its runfiles. This tends to hit the worst-case, because CD references the deployable artifacts which are the largest files in the CAS (e.g. docker container image layers). So we are using a fork of Bazel with Ed's patches.

EdSchouten:
Hi there! Let me try to respond to all of your messages in one go.

> @EdSchouten and others: one thing I am concerned about is ongoing maintenance. If we add this functionality to Bazel, people will come to us and report bugs or ask for features in relation to it and we currently don't have the bandwidth or use-case to staff it.
> 
> What is the level of commitment from your side to help out with maintaining the functionality?

We are strongly dependent on having this feature working, as we are using it intensively. I therefore think it's also in our interest ensure it keeps working reliably. So with regards to bugs we can surely help out. With regards to adding new features to it, I can't make any commitments to that, as it would also depend on whether those features are realistic and to a certain extent align with our use case.

> > Bazel's memory usage has increased significantly. The ActionInputMap that gets created during build may easily consume multiple gigabytes of space for a sufficiently large project.
> This should be reduced significantly by https://github.com/bazelbuild/bazel/pull/17120

No, this is different as far as I know. The ActionInputMap is specific to BwoB bookkeeping, whereas the PR you linked was about reducing memory usage while computing Merkle trees of actions when doing cache lookups.

> Right, my suggestion above was to improve Bazel's UX to give users a way to download the files they need instead of everything.

The point is that users don't always know what they need. For example, in our case it's not uncommon for people to launch GDB/LLDB sessions against some kind of executable that they built. How would a user know which files under bazel-bin/ need to be present to do that?

- On macOS you need to make sure to download the dSYM directories.
- What if some parts of the executable consist of generated code? Then in that case you may also need to have some of that generated code available locally to generate proper source listings. But you also don't need to download all of it. Only the parts that are relevant to the stack frames you're debugging.

Using a FUSE file system keeps those kinds of use cases working as they otherwise would, without requiring that the user has to jump through additional hoops.

meisterT:
Thanks for your reply, @EdSchouten!

> _The point is that users don't always know what they need._

I think this is the strongest argument in favor of this proposal. Do you have data on how often users actually access intermediate files?

I am interested to learn about more scenarios from you and others to see whether we can improve the UX for BwoB - which is a useful thing to do even if it might not solve all the pain points listed in your proposal.

My understanding is that your client daemon is Linux only, right? Do you plan to extend it to other platforms?

I would also like to hear from other remote execution service owners (cc @bazelbuild/remote-execution and @brentleyjones, also cc @jmmv who commented on the PR) whether they would implement a similar daemon once this proposal is implemented - and whether they would like to see any changes to the proposal.

brentleyjones:
> The point is that users don't always know what they need.

This partially falls under "Work seamlessly with IDEs" in https://github.com/bazelbuild/bazel/issues/6862. The new regex stuff helps, for sure, but it can be a pain to get right (I've fixed many bugs in rules_xcodeproj around this), and the requirements can change over time making this hard to keep right. Also, if a user doesn't need those files every session, they are unnecessarily downloading them, when instead it would be better to download them on demand.

Also, if someone isn't using an IDE, or tooling that gets all of those flags right, then they are currently out of luck.

> I would also like to hear from other remote execution service owners whether they would implement a similar daemon once this proposal is implemented.

When I was testing `bb_clientd` back when the implementation PR first appeared, the performance wasn't great for me. I'm using macOS and the FUSE performance was suboptimal. And by this I mean that a normal build was slower using this than using BwtB. I haven't tested it again since, but I've been hesitant because of the reliance on FUSE. AFAIK though, this proposal works with any REAPI compliant service, so we wouldn't need to implement our own daemon, and thus wouldn't want to block this proposal.

EdSchouten:
> My understanding is that your client daemon is Linux only, right? Do you plan to extend it to other platforms?

It also works on macOS, if you make sure you get the stars to align. bb_clientd supports both FUSE and NFSv4. This means that on macOS, you can either use it in combination with macFUSE or with the NFS client that's part of the kernel. Both approaches have their shortcomings, though:

- As mentioned above, macFUSE has some bugs that causes it to lock up. People who run into these issues will need to work with the maintainer of macFUSE to get those fixed.
- NFSv4 support only works reasonably well on macOS Ventura 13.3.

> I would also like to hear from other remote execution service owners (cc https://github.com/orgs/bazelbuild/teams/remote-execution and @brentleyjones, also cc @jmmv who commented on the PR) whether they would implement a similar daemon once this proposal is implemented - and whether they would like to see any changes to the proposal.

Note that bb_clientd is part of the Buildbarn project, but not coupled against it. The goal is that you can also use it in combination with any other remote execution service.

coeuvre:
> The ActionInputMap is specific to BwoB bookkeeping, 

Can you elaborate more on this? I don't think ActionInputMap is specific to BwoB, it's used for all the mode. Memory wise, I believe BwoB should be the same comparing to other modes.

qc-benjamin:
We felt the need to give the feedback that this feature is significant to Qualcomm for much the same reasons as have been brought up by others and we would much prefer it be a mainline feature rather than maintained in several unofficial forks.

The feature especially shines when building large monorepos with a wide net of targets (i.e. `//...`)  this can then be performed once and then continue the CI pipeline and/or the developer workflow transparently with regard to whether the build was done locally or remotely.

In such situations remote_download_all is unfeasible for obvious reasons and listing all inputs possibly required by latter steps in the build is also unrealistic.

We also feel the need to clarify that while bb-clientd is a good implementation of something that consumes the remote output service apis and most users of bb-clientd are likely to be using buildbarn. The actual pull request is independent of bb-clientd and bb-clientd itself is independent of buildbarn.

I.e. any REAPI compatible environment would benefit from the improved traffic flow from using bb-clientd and the remote output service can be consumed by things other than bb-clientd. It could for example be used to integrate with some other virtual filesystem. Without the information communicated by the remote output service this information becomes lost inside of bazel.

lberki:
From this thread, it looks like that there is a *lot* of demand for the output tree to be on a FUSE file system of some sort. With that in mind, I'm fine with marking this design doc as approved and merging https://github.com/bazelbuild/bazel/pull/12823 as long as @EdSchouten (or someone else, but he is the most obvious candidate) can commit to bringing it up to the standards of the code base of Bazel and putting it behind an experimental flag. WDYT?

lberki:
Friendly ping! I'd like to decide one way or the other; to reiterate: I'm fine with merging #12823 as long as someone can commit to rebasing it and bringing it up to our quality standards.

lberki:
Yeah, I don't think dropping a proposal needs to mark the "final death" of said proposal, but it's a pretty strong signal.

I wanted to keep this open for a while because @EdSchouten has invested a lot of energy into it and IIUC they maintain a fork of Bazel with that patch anyway, so I thought it would be a net win for them to not have to maintain their patch at the cost of some upfront investment.

lberki:
I'll keep this open for a while (say, a month, assuming I actually manage to remember the deadline) to signal that we are looking for volunteers and if there are none, I'll mark it as "dropped".

lberki:
At the last moment, @stagnation and @Gormo stepped up. So I'll update this pull request to say "move to approved" and will ask @coeuvre to review it since he's the most likely candidate for the eventual code review.

coeuvre (approved):

PR #319, opened 2023-06-13
Mark the credential helpers proposal as implemented.

tjgq:


lberki:
Mind also moving it to the right category in `README.md`?

tjgq:
> Mind also moving it to the right category in `README.md`?

Done.

lberki (approved):

PR #320, opened 2023-06-14
Remove mention on inline macro variant of `platform_data`

katre:


katre:
Based on feedback from https://github.com/bazelbuild/bazel/discussions/18628#discussioncomment-6143247

gregestren (approved):

PR #330, opened 2023-08-01
Mark "Bazel Proto Rules" as Under review

alexeagle:
We have some likely funding for this effort from the Protobuf team, so it's time to get it ready for implementation.


lberki (approved):

gregestren:
@comius @alexeagle any reason not to merge?

alexeagle:
I don't see a reason, thanks

PR #332, opened 2023-08-02
Add a table for documents that are not design docs.

lberki:


meisterT (approved):

lberki:
I'll also wait for @comius and @aiuto so that they can chime in if they want.

README.md
```diff
@@ -1,9 +1,12 @@
...
+# Bazel Documents
 
-This is an index of all proposals and design documents for
-[Bazel](https://bazel.build).
+This is an index of all documents for [Bazel](https://bazel.build).
 
-## Workflow
+There are two kinds:
+* Design docs that are subject to the workflow described below
+* Live documents that are constantly updated
```
comius, line 7: super nit: are any of this really "constantly updated", I'd prefer to be more truthful and skip the part after "that" ;)

README.md
```diff
@@ -14,7 +17,15 @@ list.
 
 ![State diagram of the design process](images/states.png)
 
-# Index
+# Live documents
+
+| Last updated | Title                                                                                                                                                     | Author(s) alias                               |
```
comius, line 22: I'd prefer to have also a "state" column, but then we need some agreement what that is.
I'd say: "Draft"/"Under review"/"Reviewed"/"Outdated"/"Replaced by"

lberki: I had the exact train of thought; I added a "status" field, thought a bit about what kind of statuses there should be, couldn't make up my mind, decided that it would just be subject to bikeshedding and thus removed it. We can always add one later.

lberki:
Submitting; it looks like @comius broadly agrees.

PR #339, opened 2023-08-08
Add "Standard exec configurations" to review

gregestren:
@katre  already reviewed this, and I've started implementing. But we didn't share this publicly. Doing so now to update the community and solicit wider feedback.

Note this doc is generic in its ideas but  focused on Google's repo as an initial test of "cleaning up a repo". 

Other orgs would need to consider this independently based on their own patterns. Some of the cleanup can't generalize to other orgs because of that.


gregestren:
Also sharing because this has some cross-over with [Platform-based flags](https://github.com/bazelbuild/proposals/blob/main/designs/2023-06-08-platform-based-flags.md).

lberki (approved):

katre (approved):

PR #346, opened 2023-09-14
Drop duplicate proposal which is a 404

alexeagle:
Duplicate of "Extending Bazel rules" by @comius I believe


comius:
cc @lberki I'd be interested in reading this if you can find it.

lberki (approved):

PR #349, opened 2023-10-09
Fix some YAML errors in proposals

saraadams:
In some of the proposals, the header with the metadata on the proposal does not render correctly on GitHub. Example:

```
Error in user YAML: (<unknown>): found character that cannot start any token while scanning for the next token at line 5 column 5
```

This seems to be caused by entries starting with an `@` symbol. This change fixes that.


lberki (approved):
Thanks!

PR #352, opened 2023-10-18
Add Bazel Ruleset GOVERNANCE | MECHANICS

comius:


lberki:
Mind a adding a "Status" field to the document? It looks like from the comment threads that there are disagreements about the desired governance structure, so I think it's too early to codify this as gospel.

alexeagle:
@comius is this on your agenda for our Authoring Rulesets bof?

comius:
> Mind adding a "Status" field to the document? It looks like from the comment threads that there are disagreements about the desired governance structure, so I think it's too early to codify this as gospel.

Done

lberki:
Far be it from me to nitpick, but is the ALL CAPS NECESSARY?

lberki (approved):

comius:
> @comius is this on your agenda for our Authoring Rulesets bof?

Sure. Let's see first if I can address the comments, so that I get more familiar with the domain. (I'm not the author).

@alexeagle do you think you'd be a good primary reviewer? Or could you suggest somebody else?

alexeagle:
Sure, I already did a preliminary review as Radhika was composing this, I'm happy to be a reviewer of record, thanks.

lberki (approved):

PR #355, opened 2023-10-25
add aspect pruning proposal

aiuto:
[Faster aspects: Pruning on the way down](https://docs.google.com/document/d/1Md4BX2v-VMV-qNmc-wdeImTCdBK9BRF6YNktmOBVVVY/edit)


lberki:
Isn't this something better suited for the "draft" category?

aiuto:
> Isn't this something better suited for the "draft" category?

Yes. Done. And the partner doc added.

"Live documents" sounds confusingly similar to "draft" to me. Or at least, it seems more "I'm still changing" than under review or approved.

lberki:
I got it; maybe we should add an explanation in README.md that a "live document" is one that's not subject to the usual review workflow because it's not a design doc.

lberki (approved):

PR #359, opened 2023-12-06
Exec platform scoped spawn strategies

Silic0nS0ldier:
Proposal to make spawn strategies aware of execution platforms via the existing strategy CLI flags.

## Goals

* Keep the scope small, such that this proposal is easier to review and implement.
* No new concepts, build off what already exists.
* Make it easy for existing workflows to take advantage (e.g. `--config remote` to enable remote execution in a build).

## Prior Proposals

Solutions to this problem space have been proposed twice before (that I know of).

- 2019-08-01 - [2019-06-25-platforms-strategies.md](https://github.com/bazelbuild/proposals/blob/1032d7d8fbe0e518e88482e7c532eb409ef6f1ac/designs/2019-06-25-platforms-strategies.md) by @katre
- 2023-03-31 - [Execution Platforms vs. Strategies](https://docs.google.com/document/d/1U9HzdDmtRnm244CaRM6JV-q2408mbNODAMewcGjnnbM) by @katre and @aragos

## Potential Follow-ups

- `--remote_local_fallback` is unresolved by this proposal as it does not rerun spawn strategy selection, however all the pieces required to support such behaviour are in place. 
  Supporting this will require changing behaviour (new flag or major release required) and goes beyond the scope of this proposal (mnemonics, description matching regex, and so on would be considered in addition to execution platform scoping).
  It would make sense to see this addressed in a separate proposal.

PR #361, opened 2023-12-07
`--remote_local_fallback` Respects Strategy Declarations

Silic0nS0ldier:
This is a follow up to #359 proposing to make `--remote_local_fallback` fallback to a freshly picked non-remote spawn strategy (if there are candidates left) rather than `local`.

## Related Issues

- https://github.com/bazelbuild/bazel/issues/7202
- https://github.com/bazelbuild/bazel/issues/15519

### DIFF
--- a/designs/2023-12-07-remote-local-fallback-respects-declarations.md
+++ b/designs/2023-12-07-remote-local-fallback-respects-declarations.md
@@ -1,6 +1,6 @@
 ---
 created: 2023-12-07
-last updated: 2023-12-07
+last updated: 2023-12-08
 status: Draft
 reviewers:
   - TBD
@@ -12,7 +12,7 @@ authors:
 # Abstract
 
 The flag `--remote_local_fallback` makes it possible for remotely executed action spawns that fail to fallback to the `local` spawn strategy.
-This proposal seeks to treat spawn retries behave identically to actions tagged with `no-remote`.
+This proposal seeks to treat spawn retries identically to actions tagged with `no-remote`.
 
 # Background
 
@@ -25,7 +25,7 @@ e.g.
 build --strategy=Foo=remote
 ```
 
-Requiring that an action be only spawned locally may reflect a technical limitation.
+Requiring that an action be only spawned remotely may reflect a technical limitation.
 Such as the remote offering a Windows environment while Bazel itself is running on a macOS machine.
 The Windows binaries cannot be reasonably expected to work on macOS, and compatibility layers to support such a scenario are unlikely to provide identical behaviour.
 
@@ -40,11 +40,13 @@ Even in a context where remote and local environments are identical, `local` may
 
 # Proposal
 
-A new flag `--incompatible_strict_remote_local_fallback` will be introduced which when flipped makes `--remote_local_fallback` rerun spawn strategy with `remote` ignored.
+A new flag `--incompatible_strict_remote_local_fallback` will be introduced which when flipped makes `--remote_local_fallback` rerun spawn strategy selection with `remote` ignored.
 
 This means;
 * All registered strategies along with their filters are considered (`--spawn_strategy`, `--strategy`, `--strategy_regexp`, etc).
 * If a `remote` spawned action has no local fallback, no attempt to spawn locally is made and the build fails.
+
+To avoid confusion, the `--remote_local_fallback_strategy` flag (currently marked as a deprecated no-op, ([#7480](https://github.com/bazelbuild/bazel/issues/7480))) will be removed.
 
 # Backward-compatibility
 
--- a/designs/2023-12-07-remote-local-fallback-respects-declarations.md
+++ b/designs/2023-12-07-remote-local-fallback-respects-declarations.md
@@ -1,9 +1,9 @@
 ---
 created: 2023-12-07
-last updated: 2023-12-08
-status: Draft
+last updated: 2023-12-12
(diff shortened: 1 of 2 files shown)


designs/2023-12-07-remote-local-fallback-respects-declarations.md
```diff
@@ -0,0 +1,51 @@
...
+* `sandboxed` may provide an environment closer to that of the remote.
+
+## Related Issues
+
+* [remote: remove local fallback for remote execution](https://github.com/bazelbuild/bazel/issues/7202)
+* [Make --remote_local_fallback honor --spawn_strategy](https://github.com/bazelbuild/bazel/issues/15519)
+
+# Proposal
+
+A new flag `--incompatible_strict_remote_local_fallback` will be introduced which when flipped makes `--remote_local_fallback` rerun spawn strategy with `remote` ignored.
```
tjgq, line 43: FWIW, it's not even obvious to me that we need to guard this change behind a flag; it might be the case that everyone already regards the current behavior as a bug, and we can retroactively fix it without causing problems for anybody.

designs/2023-12-07-remote-local-fallback-respects-declarations.md
```diff
@@ -0,0 +1,51 @@
...
+  - TBD
+title: `--remote_local_fallback` Respects Strategy Declarations
+authors:
+  - Silic0nS0ldier
+---
+
+# Abstract
+
+The flag `--remote_local_fallback` makes it possible for remotely executed action spawns that fail to fallback to the `local` spawn strategy.
+This proposal seeks to treat spawn retries behave identically to actions tagged with `no-remote`.
```
katre, line 15: I like the idea but remove the "behave" and I think this reads much cleaner.

katre (dismissed):

katre:
> @katre are you comfortable with reviewing this also?

I am willing to, but I'm not super familiar with the workings of the remote strategy and local fallback. Possibly @tjgq is a better choice?

designs/2023-12-07-remote-local-fallback-respects-declarations.md
Silic0nS0ldier: Strictly speaking the behaviour is not a bug. `--remote_local_fallback` explicitly states remote fallback uses the `local` execution strategy. Not "a" local execution strategy.

> Whether to fall back to standalone local execution strategy if remote execution fails.

That said, I personally have no qualms about changing this without a major version increment. It would make implementing the change easier (1 PR vs. 2+ at potentially different times).

I don't think fallbacks failing to kick in when no local spawn strategy can be resolved is itself represents a breaking change either. Changing from `local` to `worker` or `sandboxed` is of more concern, while different spawn strategies shouldn't lead to different results it does happen in practise ([Hyrum's Law](https://www.hyrumslaw.com/)?).

designs/2023-12-07-remote-local-fallback-respects-declarations.md
```diff
@@ -0,0 +1,51 @@
...
+* [Make --remote_local_fallback honor --spawn_strategy](https://github.com/bazelbuild/bazel/issues/15519)
+
+# Proposal
+
+A new flag `--incompatible_strict_remote_local_fallback` will be introduced which when flipped makes `--remote_local_fallback` rerun spawn strategy with `remote` ignored.
+
+This means;
+* All registered strategies along with their filters are considered (`--spawn_strategy`, `--strategy`, `--strategy_regexp`, etc).
+* If a `remote` spawned action has no local fallback, no attempt to spawn locally is made and the build fails.
+
```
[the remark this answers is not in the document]
Silic0nS0ldier: According to documentation the flag is already a deprecated no-op. Assuming this is true, it would make sense to move it into the graveyard and tag it appropriately (is missing `OptionMetadataTag.DEPRECATED`, like a lot of deprecated options as it turns out).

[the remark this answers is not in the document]
tjgq: It turns out that it's actually not a no-op, unfortunately (see recent discussion at https://github.com/bazelbuild/bazel/issues/15519#issuecomment-1841940599).

designs/2023-12-07-remote-local-fallback-respects-declarations.md
tjgq: Okay, I am convinced that there's enough subtlety here to justify following the incompatible change process (as the text is currently proposing).

lberki:
@Silic0nS0ldier , @tjgq graciously agreed to review this one. Mind updating the reviewer in the document and changing the category to "under review"?

Silic0nS0ldier:
@lberki Sure. I'll make that change tomorrow, along with the appropriate tweaks for the not-no-op `--remote_local_fallback_strategy` flag.

designs/2023-12-07-remote-local-fallback-respects-declarations.md
Silic0nS0ldier: In light of the discovery that `--remote_local_fallback_strategy` does in fact work despite being documented a no-op, I've kept the incompatible flag. Mainly so this can be discussed in the formal review.

Silic0nS0ldier:
@lberki PR has been updated.

lberki (approved):

lberki:
Thanks, merging

PR #385, opened 2024-06-17
Adding design doc for flag propagation

aranguyen:


brentleyjones (commented):
Would this work better as a link in the README?

PR #396, opened 2024-11-22
Copy action

Silic0nS0ldier:
Proposal to introduce a builtin copy action.

## Goals

* Simplify rule development.
* Improve performance of copy-heavy toolchains (e.g. Rules JS, which accommodates NodeJS's relative import logic by copying inputs to the "bin" directory by default).
* Cut down on remote execution and BES protocol chatter caused by having lots of short-lived copy file spawns.
* Discourage optimisation anti-patterns like setting [`no-remote` and `no-cache`](https://github.com/bazelbuild/bazel-skylib/blob/5c071b5006bb9799981d04d74a28bdee2f000d4a/rules/private/copy_common.bzl#L43-L44) that can ultimately lead to greater overhead in remote builds (e.g. forcing remote files to be downloaded under `--remote_download_minimal`).


designs/2024-08-14-copy-action.md
```diff
@@ -0,0 +1,162 @@
...
+
+# Proposal
+
+Introduce a new `copy` action with the following behaviors;
+- Content hash of output is identical to input, only the name is different.
+- Realisation of output is deferred, similar to how remote spawn actions are handled.
+- Action result is empty, similar to symlink actions.
+
+The API looks like `actions.copy(in: File, out: File, path: String|None): None`.
+- `actions.copy(in, out)` is a simple same-type copy operation.
```
tjgq, line 66: Does "same type" refer to the artifact type or to the filesystem type? The two don't necessarily match: artifacts of file or directory type can be materialized as symlinks in the filesystem.

What should `actions.copy(in, out)` do if `in` and `out` are both `declare_file` (or both `declare_directory`), but `in` is materialized in the filesystem as a symlink? I think there are three possible answers:

- Copy the symlink as-is
- Indirect through the symlink and make a copy of its target
- Execution error (I like this one the least because it makes builds with the same analysis properties and input digests result in different execution behavior, depending on the state of the filesystem - a sort of non-determinism, if you will).

designs/2024-08-14-copy-action.md
```diff
@@ -0,0 +1,162 @@
...
+  Unanswered question: How are `actions.symlink` inputs handled? Update the target? Error?
+- `actions.copy(in, out, path = "foo/bar")` copies a file from the input directory.
```
tjgq, line 68: To be clear, I think you're referring to the case where `in` is a `declare_directory` and `out` is a `declare_file`? Can you work that into the sentence?

designs/2024-08-14-copy-action.md
```diff
@@ -0,0 +1,162 @@
...
+  `path` cannot be used with generic symlinks (`actions.declare_symlink`, source files??? and within tree artifacts support is later added) as Bazel does not track their target.
+
+Compared to existing solutions, `actions.copy` should;
+- Complete much more quickly, as spawn machinery is not used and no subprocess is used.
```
tjgq, line 72: An interesting question is whether the implementation is required to make an actual copy, or if hardlinks would suffice where supported.

Do you know whether hardlinks also present problems for the use cases where `actions.symlink` doesn't work?

designs/2024-08-14-copy-action.md
```diff
@@ -0,0 +1,162 @@
...
+  - ...in scenarios where clients are forbidden from uploading `ActionResult` metadata (listed outputs can be arbitrary, representing a security issue).
+  - ...when copying files generated by the remote as they must be downloaded first.
+- Implementation changes can lead to cache misses.
+- If copies are batched, odds are in most runs only 1 input will have actually changed (doing more work than necessary, becomes a performance and efficency concern once the batch hits a certain size).
+
+# Proposal
+
+Introduce a new `copy` action with the following behaviors;
+- Content hash of output is identical to input, only the name is different.
+- Realisation of output is deferred, similar to how remote spawn actions are handled.
```
tjgq, line 62: Implementation-wise this is probably the most challenging part, because none of the in-process actions (symlink, expand_template, write, etc) are currently deferred; the "build without the bytes" machinery only works for remote actions.

designs/2024-08-14-copy-action.md
```diff
@@ -0,0 +1,162 @@
...
+- Action result is empty, similar to symlink actions.
+
+The API looks like `actions.copy(in: File, out: File, path: String|None): None`.
+- `actions.copy(in, out)` is a simple same-type copy operation.
+  Unanswered question: How are `actions.symlink` inputs handled? Update the target? Error?
+- `actions.copy(in, out, path = "foo/bar")` copies a file from the input directory.
+  `path` cannot be used with generic symlinks (`actions.declare_symlink`, source files??? and within tree artifacts support is later added) as Bazel does not track their target.
```
tjgq, line 69: Using it with source artifacts should be fine (what determines whether Bazel tracks the contents is the artifact type, not the filesystem type, and source artifacts are currently always of `file` type) but then we run into the same question as above in cases where they're materialized as symlinks.

designs/2024-08-14-copy-action.md
```diff
@@ -0,0 +1,162 @@
...
+  - ...when copying files generated by the remote as they must be downloaded first.
+- Implementation changes can lead to cache misses.
+- If copies are batched, odds are in most runs only 1 input will have actually changed (doing more work than necessary, becomes a performance and efficency concern once the batch hits a certain size).
+
+# Proposal
+
+Introduce a new `copy` action with the following behaviors;
+- Content hash of output is identical to input, only the name is different.
+- Realisation of output is deferred, similar to how remote spawn actions are handled.
+- Action result is empty, similar to symlink actions.
```
tjgq, line 63: I think what you mean here is that the result of the action isn't uploaded to a remote or disk cache?

designs/2024-08-14-copy-action.md
Silic0nS0ldier: Hardlinks are probably fine for outputs, only surprising if output tree checking is turned off and a file is modified.

Copy-on-write is also an option (macOS has `clonefile`, Linux has `reflink`, Windows has something recent), though it requires the source and destination to be on the same volume, and for the filesystem to support it. Cached feature detection would be necessary for optimal performance.

designs/2024-08-14-copy-action.md
sluongng: I disagree with having an empty action result. I think it's still valuable to have one I think things might break badly if we don't have one to track the output directory life time.

designs/2024-08-14-copy-action.md
sluongng: +1 to Tiago comment.

I would prefer to see separate definitions for cases such as:
- File to File
- Dir to File
- File to Dir

I am also ok if we decide to only support File to File initially.

designs/2024-08-14-copy-action.md
sluongng: I would prefer if we start with an actual copy and add options to use hard link or copy-on-write(CoW) copy separately.

designs/2024-08-14-copy-action.md
```diff
@@ -0,0 +1,162 @@
...
+- Cut down on remote execution (TBD - fairly sure this shouldn't have an action metadata record) and BES (TBD - need to check protocols and get some traffic samples) traffic during batches of copy operations.
+
+## Example Usage
+
+### Skylib's [`copy_directory`](https://github.com/bazelbuild/bazel-skylib/blob/5c071b5006bb9799981d04d74a28bdee2f000d4a/rules/private/copy_directory_private.bzl)
+
+```starlark
+def _copy_directory_impl(ctx):
+    dst = ctx.actions.declare_directory(ctx.attr.out)
+    ctx.actions.copy(ctx.attr.src, dst)
```
sluongng, line 83: In this example, the desired output here would be a Directory which should be uploaded to remote cache CAS so that subsequent downstream actions could consume it. 

If the copy is large (i.e. Dir to Dir recursively), having an ActionResult would also save other Bazel clients from having to do the work.

---

I also think that if `File to Dir` is supported, then we should consider making it a list of files instead so we can compose the directory contents in one go.

sluongng (commented):
FWIW, this is something that we have discussed internally with BuildBuddy as these copy actions are used heavily among the rules_js, rules_oci ecosystem.

Specifically for rules_oci, the copy contents could range anywhere between a few bytes of json, to multi-gigabytes of compressed tarball container layer. Which make them quite "expensive" to schedule and execute remotely.

An alternative way of achieving this, which we are discussing internally, is for the RBE scheduler to "detect" the known copy actions and skip the executions completely by creating the ActionResult directly instead. Though, it would be much nicer if it's Bazel doing these "short-circuit" executions instead of our server.

designs/2024-08-14-copy-action.md
Silic0nS0ldier: I did a _very_ poor job of describing the requirement here, and this is after having rewritten this several times (while failing to fully recall what I was on about).

Thinking on this again now, I _think_ my past self was concerned about how remote/BES APIs would respond to a large number of copy operations (10k+) that are unbatched. That is, it is very easy to process 10k copy actions in a second (or there abouts) so keeping the number of events/calls they produce as a side effect low is important for keeping connections from becoming flooded and remote services healthy.

There is definitely a better way to phrase this requirement, I'll need to poke around at the remote/BES API surfaces to fill some knowledge gaps first I think. This may be a simple as batching (not of the actions themselves, more the side effects like BES events and remote interactions).

designs/2024-08-14-copy-action.md
Silic0nS0ldier: > Does "same type" refer to the artifact type or to the filesystem type? The two don't necessarily match: artifacts of file or directory type can be materialized as symlinks in the filesystem.

I was thinking artifact type when I wrote this, symlinks are something I've yet to figure out. It's very easy to break them, which to be fair is the case for the existing copy rules. The challenge here is that with _some_ symlinks Bazel will actually dereference them and track the targeted artifact.

Not sure what behaviour to go with here just yet. I feel like "copy-as-is" would be the least surprising option, keeps it simple.

> - Indirect through the symlink and make a copy of its target

Would make sense when in the source tree a file has been symlinked specifically so it's content can appear in multiple locations. Such cases would be ideally served by hardlinks (multiple canonical handles, same file) but that's not something version control systems keep track of.

Symlinks to a directory ramps up the complexity. They can be used to present the same content in multiple locations, but they also have more novel uses like in `pnpm`.

Additionally symlinks can point to _anything_, including other symlinks (e.g. `symlink-a -> symlink-b -> file`, `symlink-a -> symlink-b -> symlink-a`). Solving for that in the copy implementation doesn't sound like much fun.

A case could also be made for reading the symlink target then creating the new "copy" with an adjusted target path.

Opting to _not_ do anything special for symlinks may be the best option, can always do the special niche operation (e.g. copying the file a symlink points to) in a separate action.

designs/2024-08-14-copy-action.md
brentleyjones: Btw, executable hardlinks have an issue with macOS Gatekeeper killing the wrong executables:
https://developer.apple.com/forums/thread/663456.

Silic0nS0ldier:
Work on this proposal had stalled as I needed a prototype to validate the design. AI has made significant strides since late 2024, so I've been able to (now) quickly put one together and give this proposal a much needed update.

It is now ready for another review.

sluongng:
there is a new submission here https://github.com/bazelbuild/proposals/pull/419 with a PR in Bazel as well. I recommend trading notes with @thesayyn first and potentially merging the 2 proposals into one?

thesayyn:
I'd be happy to work with @Silic0nS0ldier, i do believe that copying is still not a the way to solve this, i do believe leaving the copying to the execution strategy is the way to go since they are extremely optimized for staging contents on specific paths before execution.  Anything else we do in the name of copying will be duplicating bytes on disk and same pollution as we have today with Copy actions that use coreutils etc.

thesayyn:
For copy specifically Bazel already maintains one node for the file, if we design the copy api in a way that duplicates the node, file on disk,  we are still not saving anything and frankly making a worse copy action that what we have today. 

We really need the new copy api to be as cheap as it could be, i came to conclusion that it is only possible with deferred staging (leaving the materialization to the execution strategy because they are heavily optimized for staging inputs cheaply) and skyframe node aliases (cheap clone of the File handles that carries the digest from the backing file)

This will only be beneficial if its cheap. rulesets such as rules_js, rules_oci, rules_python, rules_py etc  need to workaround the fact that the tools that are being run (nodejs, python) want a specific filesystem layout to function, site-packages, node_modules,  mtree

Silic0nS0ldier:
I did notice thesayyn/bazel#3 while updating this PR, but not the proposal till just now.

Reading over #419 (`ctx.actions.alias`), as written it is orthogonal to this copy action proposal as the on-disk representation is not an explicitly guaranteed part of that proposal.

Essentially;
- `ctx.actions.symlink` (existing): Materialise the specified target with a symlink.
- `ctx.actions.copy` (this proposal): Materialise the specified target as a normal file/directory (copy, copy-on-write, hardlink, and potentially junctions on Windows now that I think about it).
- `ctx.actions.alias` (#419): Materialise the specified target here (symlink, hardlink, copy, copy-on-write, bind-mount).

Optimisations are possible across all three (lazy materialisation, deferral to a separate output service a-la `--experimental_remote_output_service`, symlink templates a-la `--remote_download_symlink_template`, etc).

The primary motivation for `ctx.actions.copy` is that the materialisation won't use symlinks which affect runtimes such as NodeJS (require/import machinery dereferences symlinks by default). `ctx.actions.alias` does not satisfy this requirement out-of-the-box (symlink-sensitive workloads would require a flag to be passed to Bazel).

-----

That all said, I don't believe the proposals are mutually exclusive. Sometimes we want a symlink, sometimes we want a copy, sometimes all we really need is for an artifact to show up in another location. For the latter, `ctx.actions.alias` can eek out some minor benefits compared to explicit symlinking (works across volume boundaries but has dereferencing overhead at runtime) and copying/hardlinking (no dereferencing). That is, it can pick the best possible strategy when the implementation is less of a concern.

-----

As far as the optimisations _in this proposal_, I have deliberately kept them as follow up items (not part of the proposal itself) so as to not hinder landing of the core API. The 1 must-have optimisation (to graduate the API from experimental status) IMO is deferral of source directory copying under remote execution. It _is_ a weakness vs. what spawn based copies can do, but in prototyping proved to be difficult to implement. Keeping this unspecified allows the disk-requirement gap to be closed via hardlinking and copy-on-write (which would likely be _much_ easier to backport).

Silic0nS0ldier:
Benchmarks table in the proposal (data from the prototype) is squashed under Githubs markdown renderer, so here is a VSCode rendering that's a little easier to read.

image

designs/2024-08-14-copy-action.md
Silic0nS0ldier: `File to Dir` won't be supported, at least not in _this_ proposal. There is too much ambiguity around what the file's path within the created folder will be. For it to be useful, some degree of customisation (i.e. customising subpaths for files placed in the directory) would also be necessary.

Since this comment a new action `ctx.actions.map_directory` has been implemented, which may be a better fit.

designs/2024-08-14-copy-action.md
Silic0nS0ldier: Addressed in updates.

designs/2024-08-14-copy-action.md
Silic0nS0ldier: Complexity around this has been noted in the updated proposal. I've intentionally kept optimising source artifacts out to afford some flexibility (for example copy-on-write/hardlinking reduce the storage requirements, are easier to implement and backport; deferring materialisation of source artifact copies can be implemented out-of-band).

designs/2024-08-14-copy-action.md
Silic0nS0ldier: macOS Gatekeeper issues around executable hardlinks noted in the updates. If/when hardlinking support is implemented, a way to opt-out (e.g. flag) should be included.

designs/2024-08-14-copy-action.md
Silic0nS0ldier: Clarified in proposal updates, but to answer the question the following are supported with `path` usage;

- Dir -> File, copying a file out of a directory.
- Dir -> Dir, copying a directory out of the input directory.

File -> Dir is intentionally not supported, `ctx.actions.copy` is the wrong API for the job. More details in https://github.com/bazelbuild/proposals/pull/396#discussion_r3634910299.

designs/2024-08-14-copy-action.md
Silic0nS0ldier: The behaviour here has been clarified in the updated proposal. Some changes vs. my last comment driven by prototyping discoveries. To recap;

- `declare_directory` -> `declare_directory`
- `declare_file` -> `declare_file`
- `declare_symlink` -> `declare_symlink`

If a `declare_directory` or `declare_file` input is provided that was fulfilled by `ctx.actions.symlink`, the symlink is dereferenced to perform the copy. I'm not convinced this is the right decision, it's a tricky spot as symlink fulfilled artifacts are a little special in their own right (materialised symlinks fulfilling directories or files do not have a fixed target path). Whatever the final outcome, the goal is to adhere to the [principle of least astonishment](https://en.wikipedia.org/wiki/Principle_of_least_astonishment).

fmeum:
The current behavior of  `ctx.actions.symlink(target_file = ...)` is particularly bad because how the output is observed on disk or by consuming actions already depends on the context:
1) On Windows, the on-disk representation may be a true copy.
2) With BwoB, the symlink action already creates the symlink in the in-memory file system and could (this is just a missed optimization) create the symlink on disk only when explicitly requested.
3) With remote execution, the artifact is indistinguishable from a true copy to the remote executor on the level of the REAPI protocol. Remote executors already use a bunch of optimizations for efficient staging, which may involve CoW, FUSE filesystems or hardlinks.
4) With local execution on Unix, the artifact is a symlink to the original file.
5) With sandboxed execution on Unix, the artifact is a symlink to a symlink to the original file.

If a tool cares about the particular symlink representation, the above already effectively force it to use `ctx.actions.symlink(target_path = ...)` instead, which behaves consistently across all platforms and execution strategies.
So whatever new mechanism is introduced, perhaps its time to deprecate `ctx.actions.symlink(target_file = ...)` in favor of it? This particular parameter just doesn't serve any purpose that isn't better served by something else.

The crucial point of any new mechanism should be that it explicitly *doesn't* prescribe the materialization strategy for the output as an input to actions, except that it guarantees one that makes the output independent of the input for all practical purposes (in particular, it must *not* be a symlink to it). Whether hardlinks are allowed is a subtle question since hardlinks change the `ctime` of the input file. Anything beyond that (CoW etc.) would be an implementation detail of Bazel and the local or remote executors. The more we focus on the semantics of the new primitive, the more optimization opportunities we leave to the implementation, which is a good thing. As performance sensitive users will use BwoB anyway, the materialization on disk also doesn't matter.

I don't think that either proposal needs to focus on how Bazel internally represents these files. Having an action registered for them is already more overhead than anything else Bazel could retain about the files, so that's a natural limit for the scale at which any such action could be used.

thesayyn:
> The crucial point of any new mechanism should be that it explicitly doesn't prescribe the materialization strategy for the output

This is the key point, `ctx.actions.copy()` can't guarantee on disk presentation because we are dealing with a matrix of `3 kernel x N filesystems` + `remote execution`. Any guarantee we make about semantics of `ctx.actions.copy()/alias()` is a lie, hence my take on it is that we don't even copy anything its just a pointer to an existing file and gets a different materialization path during execution. This works flawlessly for every execution strategy, even local (without sandbox).

Any other eager materialization strategy, eg CoW/hardlink/reflink by the `CopyAction` is making a choice that prevents any further optimization down the line, eg by execution strategies. So i agree with Fabians take on this. 

> I don't think that either proposal needs to focus on how Bazel internally represents these files

I agree, its bazels internal and rulesets don't care how its tracked as long as there are no correctness issues.

designs/2024-08-14-copy-action.md
Silic0nS0ldier: Given the behaviour of `ctx.actions.symlink(target_file = ...)` described in https://github.com/bazelbuild/proposals/pull/396#issuecomment-5061810133, dereferencing looks to be the most consistent (and maybe the only possible) way to handle this scenario.

Silic0nS0ldier:
> Whether hardlinks are allowed is a subtle question since hardlinks change the ctime of the input file.

That's a point in favour of hardlinks being opt-in, if we care about that side effect. From memory EngFlow's remote execution implementation uses hardlinks to materialise inputs in executors, some other services likely do the same. Biggest concern I can see in Bazel is that a `ctime` change trips up change detection, leading Bazel to rehash source inputs unnecessarily or possibly rerun the generating action for generated inputs.

> The crucial point of any new mechanism should be that it explicitly doesn't prescribe the materialization strategy for the output as an input to actions, except that it guarantees one that makes the output independent of the input for all practical purposes (in particular, it must not be a symlink to it).

Taken in the context of #419, that would mean `ctx.actions.alias` cannot use symlinks (as that would make the output dependent on the input). Unless the scope were broadened to include REAPI changes necessary to express that a given artifact is safe to materialise with a symlink.

> Any other eager materialization strategy, eg CoW/hardlink/reflink by the `CopyAction` is making a choice that prevents any further optimization down the line, eg by execution strategies. So i agree with Fabians take on this.

I don't quite follow, or maybe I've overlooked something in the proposal. What is the concern that impacts the ability to implement further optimisations?

Silic0nS0ldier:
Small update. I've been working on a lazy download proposal (allows rules to declare non-configurable downloads that can't be eagerly triggered by `bazel cquery`, keeps bytes off host under RBE with a remote downloader) which would benefit from making artifact selection more flexible. e.g. picking an executable in a directory (like from an extracted archive) without needing to perform a copy (nor any other functionally equivalent optimisation).

If a artifact selection API were to be introduced, it would make the `path` argument in `ctx.actions.copy` redundant. Under the current prototype, copying a file out of a directory would look something like;
```starlark
foo = ctx.actions.pick_file(bar_dir, "foo")
foo_copy = ctx.actions.declare_file("foo_copy")
ctx.actions.copy(foo, foo_copy)
```

It's 3 lines vs. 2, but would avoid having 2 ways to achieve the same outcome.

thesayyn:
> I don't quite follow, or maybe I've overlooked something in the proposal. What is the concern that impacts the ability to implement further optimisations

If `ctx.actions.copy` registers an action that does the copy via reflink,clonefile,hardlink etc, then the optimization of never copying inputs is thrown out the window. If your proposal is not going to do that, that's basically what i have already proposed via lazy materialization `ctx.actions.copy` api. 

> Taken in the context of https://github.com/bazelbuild/proposals/pull/419, that would mean ctx.actions.alias cannot use symlinks (as that would make the output dependent on the input). Unless the scope were broadened to include REAPI changes necessary to express that a given artifact is safe to materialise with a symlink.

the proposal goes beyond RE, its also taking the local execution strategies into account and has the ability to (open question) express content materialization via execution_requirements so things such as nodejs can say i need something that does not use symlinks.

Silic0nS0ldier:
Lazy materialisation where possible is included under this proposal (with one caveat).

> 4. **Realisation of the output is deferred where possible.** <span id="deferred-realisation"></span>
>    When the input's content is remote-backed under Build without the Bytes, the copy completes as a metadata-only operation and the output is materialised on demand, exactly like any other remote-backed output.
>    > [!NOTE]
>    > Lazy realisation of source artifacts is out of scope for this proposal as Bazel's rewinding machinery recovers a lost artifact by re-executing its generating action. Source artifacts have no generating action, and [`ActionRewindStrategy`](https://github.com/bazelbuild/bazel/blob/5cce7794834a1531ee5fad131f11a9d4d8e66781/src/main/java/com/google/devtools/build/lib/skyframe/rewinding/ActionRewindStrategy.java) requires all lost artifacts to be derived.
>    >
>    > Related work:
>    > - [Lazily create runfiles symlink trees with BwoB](https://github.com/bazelbuild/bazel/pull/26971)
>    > - [Lazily create local symlinks with BwoB](https://github.com/bazelbuild/bazel/pull/26045)
>    > - [Add `--file_write_strategy` to Bazel](https://github.com/bazelbuild/bazel/pull/24921)

---

> the proposal goes beyond RE, its also taking the local execution strategies into account and has the ability to (open question) express content materialization via execution_requirements so things such as nodejs can say i need something that does not use symlinks.

This is a part of the `copy.actions.alias` proposal I am concerned about. Local execution materialising via symlinks and remote execution materialising via hardlinks/CoW is essentially the same behaviour as the existing `ctx.actions.symlink` API (when the destination is not a `declare_symlink`). Runtimes like NodeJS are known to behave differently in their import resolution when symlinks are involved, but that is not a unique constraint. The same issue can easily come up in other places, and when local and remote diverge it leads to rules that do not behave consistently.

I don't think that `execution_requirements` is an appropriate answer to this problem, it just pushes the responsibility onto rule authors (in the implementation) and users (there will be cases where the default materialisation under local execution is not the desired behaviour).

oga35767-eng (approved):

oga35767-eng:
👍

PR #411, opened 2025-06-19
add doc for sh_toolchain for run_shell.

jacky8hyf:
https://github.com/bazelbuild/bazel/issues/26326

https://github.com/bazelbuild/bazel/issues/5265

### DIFF
--- a/designs/2025-06-18-sh-toolchain-for-run-shell.md
+++ b/designs/2025-06-18-sh-toolchain-for-run-shell.md
@@ -1,0 +1,187 @@
+---
+created: 2025-06-18
+last updated: 2025-06-18
+status: To be reviewed
+reviewers:
+  - fmeum
+title: Template for proposals
+authors:
+  - jacky8hyf
+---
+
+
+# Abstract
+
+`run_shell` actions should respect registered `sh_toolchain()`.
+
+A shell binary, as well as all tools available in `PATH`, should be declared in
+the `sh_toolchain()`. Then, all actions created by `ctx.actions.run_shell()`
+should use the given shell and `PATH` to resolve commands.
+
+# Background
+
+As of Bazel 8.2.1, `ctx.actions.run_shell` on a Linux machine does not respect
+the `sh_toolchain()` registered. In addition, even with
+`--incompatible_strict_action_env`, `PATH` is only frozen to a restrictive list
+like `/usr/local/bin:/usr/local/sbin:/usr/bin:/usr/sbin:/bin:/sbin:.`. The
+actions can still access tools from the host in `/usr/bin`, for example.
+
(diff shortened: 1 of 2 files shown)


designs/2025-06-18-sh-toolchain-for-run-shell.md
```diff
@@ -0,0 +1,187 @@
+---
+created: 2025-06-18
+last updated: 2025-06-18
+status: To be reviewed
+reviewers:
+  - fmeum
```
fmeum, line 6: @ahermann @meteorcloudy @comius Who is up for being a reviewer?

malt3: I can take a look from the Tweag side (and ensure that Nix support is possible with this design). (cc @aherrmann)

jacky8hyf: Added @malt3 ; leaving this open in case more reviewers need to be added

jacky8hyf:
Hello, may I ask if there are any updates to this? Is there anything else I should do to get this proposal merged?

fmeum:
> Hello, may I ask if there are any updates to this? Is there anything else I should do to get this proposal merged?

Since this proposal is in Markdown, I think that we are supposed to discuss it via this PR (I may be wrong though, I've mostly used Google Docs so far). I simply haven't been able to think this through yet, but I promise to do that now. Feel free to ping me again in a week if I haven't kept that promise. :-)

dzbarsky:
Would this proposal also encompass controlling the shebang? On OSX, a shebang line such as `!/usr/bin/env bash` results in silently dropping any env vars prefixed with DYLD_ due to [SIP](https://support.apple.com/en-us/102149).

I am actually interested in a slightly different problem (controlling the wrapper/launcher scripts output by rulesets such as rules_java) but as @fmeum suggested, it may make sense to align those with `run_shell`

jacky8hyf:
> Would this proposal also encompass controlling the shebang? On OSX, a shebang line such as `!/usr/bin/env bash` results in silently dropping any env vars prefixed with DYLD_ due to [SIP](https://support.apple.com/en-us/102149).
> 
> I am actually interested in a slightly different problem (controlling the wrapper/launcher scripts output by rulesets such as rules_java) but as @fmeum suggested, it may make sense to align those with `run_shell`

Personally, I think it is somewhat an orthogonal problem. This proposal focuses on having `run_shell()` to use `sh_toolchain()`. Controlling the shebang is probably what the `sh_toolchain()` itself should do, so it probably would fit as a feature request to https://github.com/bazelbuild/rules_shell/issues. I agree that if the sh_toolchain() does have the ability to control the shebang, then `run_shell()` should respect the same shebang.

lberki (approved):

PR #416, opened 2026-04-14
Add "BCR Module Indexing"

jordan-bonser:
Reviewed-by: <jordan.bonser@arm.com> # gatekeeper

### DIFF
--- a/designs/2026-04-14-bcr-module-indexing.md
+++ b/designs/2026-04-14-bcr-module-indexing.md
@@ -1,0 +1,119 @@
+---
+created: 2026-04-14
+last updated: 2026-04-14
+status: Draft
+reviewers:
+  - meteorcloudy
+title: BCR Module indexing
+authors:
+  - jordan-bonser
+---
+
+# Abstract
+
+Currently the only way to discover bazel modules in the Bazel Central Registry is to use the website(https://registry.bazel.build/).
+
+For privately hosted registries there doesn't appear to be any way to do this, and especially not from the Bazel CLI.
+
+This proposal aims to add indexing at the registry level, so all registries can have their modules be discoverable/enumerable.
+
+# Background
+
+The Bazel Central Registry is currently implemented as a static HTTP-hosted repository where each module is stored under a directory corresponding to its name. While this design keeps the hosting model simple, it lacks a registry index describing the set of available modules.
+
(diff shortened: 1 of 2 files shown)


designs/2026-04-14-bcr-module-indexing.md
```diff
@@ -0,0 +1,119 @@
...
+
+# Background
+
+The Bazel Central Registry is currently implemented as a static HTTP-hosted repository where each module is stored under a directory corresponding to its name. While this design keeps the hosting model simple, it lacks a registry index describing the set of available modules.
+
+This means unless you know a module exists in a registry and what its exact name is, it is difficult to know what modules are available.
+
+# Proposal
+
+I propose we add an `index.json` under the `modules` directory of the registry, which will list the modules that exist in the registry.
```
Wyverald, line 28: IMO we should put this file in the root directory -- it should be possible for a module named `index.json` to exist.

designs/2026-04-14-bcr-module-indexing.md
```diff
@@ -0,0 +1,119 @@
...
+"@private_llvm" (https://my_private_registry.com)
+```
+
+This functionality would make use of the `index.json` to perform the lookup.
+
+## Adding a Module
+
+As part of adding a module to the BCR, there needs to be a mechanism to ensure that a PR to add a module also has the addition of the module to the `index.json`. One approach to this would be to add logic to a script ran as part of the PR Pipeline, that would ensure the new module exists in the index. The `bcr_validate.py` may be a good candidate for this logic.
+
+## Handling Merge Conflicts
```
Wyverald, line 73: this is my biggest concern with a proposal like this, so thanks for devoting a sizeable chunk of text to it. I have just one question regarding the merge driver etc setup: would it allow multiple PRs touching `index.json` to be submitted without the PR authors needing to sync and resolve conflicts manually?

Say users A and B both submit a PR adding a new module, and both PRs change `index.json` accordingly. Both PRs are approved at around the same time. A's PR goes in first. When B's PR is about to be submitted, does B get an error, or does the PR submission process automatically run the merge driver? If B does get an error, can they somehow run the merge driver using GitHub's web UI, or will they actually have to resort to using the git CLI themselves?

lberki (approved):
Approving because the discussion about a design doc is not supposed to happen on this review but on a discussion or issue thread.

PR #417, opened 2026-05-07
Add RunfilesGroupInfo proposal

malt3:
Note: we don't have to go through the full proposal process. I'm mostly interested in the opinions of the community.
For this provider to be useful, other language rules such as rules_java and rules_python have to support it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,6 +58,7 @@ stateDiagram-v2
 
 | Last updated | Title                                                                                                                           | Author(s) alias                          | Category |
 |--------------|---------------------------------------------------------------------------------------------------------------------------------|------------------------------------------|----------|
+|   2026-05-04 | [RunfilesGroupInfo: a protocol for structured runfiles](https://github.com/hermeticbuild/rules_runfiles_group) | [@malt3](https://github.com/malt3) | Starlark Rules, Packaging |
 |   2026-04-14 | [BCR Module indexing](designs/2026-04-14-bcr-module-indexing.md) | [@jordan-bonser](https://github.com/jordan-bonser) | External Repositories |
 |   2026-03-03 | [Hierarchical constraint settings](https://docs.google.com/document/d/1Lq-0aPBOT0cat1wcFbnPmurbyo05NpZXu0uGq12H_HE/edit?usp=sharing) | [@fmeum](https://github.com/fmeum) | Configurability |
 |   2025-11-09 | [Starlark type system design](https://docs.google.com/document/d/17bGpij0gPWJFbxHLKTTwfKj8dAldD1kr8MYAa0zMUMs/edit?usp=sharing)   | [@brandjon](https://github.com/brandjon)  | Starlark |


designs/2026-05-04-runfiles-group-protocol.md
```diff
@@ -0,0 +1,251 @@
...
+
+A **completeness invariant** must hold: the union of all groups must equal `DefaultInfo.default_runfiles.files` exactly. No files may be missing and no extra files may be added. This ensures that consumers can rely on groups as a complete partitioning of runfiles. The `runfiles_group_analysis_test` rule (described below) validates this invariant.
+
+## RunfilesGroupMetadataInfo
+
+`RunfilesGroupMetadataInfo` carries per-group metadata that controls ordering and merge behavior:
+
+| Field | Type | Default | Description |
+|-------|------|---------|-------------|
+| `rank` | int | 0 | Partial ordering key. Lower rank = earlier in the output. Groups at different ranks are never merged. |
```
fmeum: Is this rank relevant only among the groups advertised by a single target? If not, different rules would need more guidance on which range to pick values in.

malt3: You may want to support merging runfiles groups for a binary that itself has data dependencies with their own RunfilesGroupInfo.
It could be useful to add some guidance on ranks (something like: -10 is for files that are shared by all targets of a rules like the interpreter, Rank -5 is for third party deps, etc.).

rickeylev:
malt3 pointed me as his proposal.

FWIW, in rules_python, we are doing something similar with PyExecutableInfo -- it has several fields that partition up the runfiles into different groups to allow people to repackage them in different ways. Its bespoke to rule_python, though.

> At most one group may have [executable_group] set to True.

Why is this? The case I'm thinking is: say you want to package multiple binaries. They significantly overlap in their dependencies. e.g. two python programs that both use torch/cuda, or a thick-client+server setup.

> RunfilesGroupInfo(
>        interpreter = depset([interpreter_binary] + interpreter_files),

Was the RHS meant to be a ctx.runfiles object instead of depset? Elsewhere it says that the sum of RunfilesGroupInfo should equal DefaultInfo.runfiles, however, Runfiles.symlinks and Runfiles.root_symlinks aren't expressible using plain depsets.

> merge_metadata() -- Dict-merges metadata from multiple sources (per-key last-wins).

It says that to_list() isn't called, but how can it ensure per-key last-wins semantics without doing so?


Questions:
* It's clear e.g. foo_binary provides RGI, but what about foo_library? One part says "_library rules (optionally propagate groups upward)", but how? lib.merge_to_limit? But as a library, what limit does it care about?
* The parts about "transforms" is rather light -- what are these transforms? Are they functions? Or a struct describing specific supported things to do?

malt3:
>At most one group may have [executable_group] set to True.
Why is this? The case I'm thinking is: say you want to package multiple binaries. They significantly overlap in their dependencies. e.g. two python programs that both use torch/cuda, or a thick-client+server setup.

I think we agree. What I'm trying to say is this: A single *_binary should only set this field to `True` for a single group to signal to the downstream packaging rule that it's own entrypoint should be added to this group.

If you package multiple binaries, each may have a single group where this field is `True`. Does that make sense?

malt3:
> Was the RHS meant to be a ctx.runfiles object instead of depset? Elsewhere it says that the sum of RunfilesGroupInfo should equal DefaultInfo.runfiles, however, Runfiles.symlinks and Runfiles.root_symlinks aren't expressible using plain depsets.

Currently, `RunfilesGroupInfo` only handles the files depset. The README of `rules_runfiles_group` currently has this wording: "When a group has metadata.executable_group == True, the packager should add the executable file, runfiles symlinks, repo mapping manifest, and other supporting files for the main entrypoint to that group's files.". 

That being said, I think you bring up a good point: `RunfilesGroupInfo` should actually define groups that support merging the depset of files, along with the depset of runfiles symlinks and the depset of runfiles root symlinks.
I'm convinced that we can do that - the current version of the module is very clearly not yet meant to be stable.

EDIT: rules_runfiles_group was updated to incorporate all parts of runfiles, including symlinks.

malt3:
> It says that to_list() isn't called, but how can it ensure per-key last-wins semantics without doing so?

This is about the additional, optional metadata provider, which doesn't handle depsets.
The metadata provider looks like this in practice:

```starlark

metadata_info = struct(
  group_a = struct(rank = -5, do_not_merge = True, weight = 1000, executable_group = False),
  group_b = struct(rank = -4, do_not_merge = False, weight = 50, executable_group = False),
)

other_metadata_info = struct(
  group_c = struct(rank = 2, do_not_merge = False, weight = 1, executable_group = True),
)

```

malt3:
> It's clear e.g. foo_binary provides RGI, but what about foo_library? One part says "_library rules (optionally propagate groups upward)", but how? lib.merge_to_limit? But as a library, what limit does it care about?

The foo_binary and foo_library rules don't (and cannot) know about any limits to the amount of groups that a packager supports.
`lib.merge_to_limit` is a helper that a packager may use to merge runfiles groups it is given.
[The `rules_runfiles_group` README does a better job of giving clear guidance](https://github.com/hermeticbuild/rules_runfiles_group#creating-groups) for `foo_library`. As a terse summary:

* if you have a `data` attribute in a _library or _binary rule, you should handle the runfiles it provides. If the target you see in `data` supports RGI, simply forward the RGI by merging it with your own and returning it. If the target doesn't support RGI, you must add the runfiles to a group. Note: the guidance on how to handle this case should be improved to ensure that the same group name is chosen, no matter who has a `data` dep on a target that doesn't support RGI. This way, multiple data deps on the same target will get deduplicated automatically.

> The parts about "transforms" is rather light -- what are these transforms? Are they functions? Or a struct describing specific supported things to do?

The README of `rules_runfiles_group` hopefully explains that better. They are functions and they implement a clear interface. The idea is to pass in rgi and the metadata provider (if it exists) and to return a new RGI and metadata from the function.

malt3:
Note: I removed the (mostly duplicated) text in this proposal document in favour of a [simple link to the README of the Starlark rules](https://github.com/hermeticbuild/rules_runfiles_group). This way, there's only a single doc that is always up to date.

malt3:
Closing in favour of a GitHub discussion: https://github.com/bazelbuild/bazel/discussions/29510

Technically, this doesn't require a formal proposal since we can add support in each language rule and don't have to get any changes merged in bazelbuild/bazel for this.

PR #419, opened 2026-07-21
Proposal:  add support for artifact aliasing

thesayyn:
This document proposes artifact aliasing in Bazel, allowing content to be addressable under multiple exec paths without copying. It details the API, semantics, and implications for backward compatibility and laydown strategies.

### DIFF
--- a/designs/2026-07-20-artifact-alias.md
+++ b/designs/2026-07-20-artifact-alias.md
@@ -193,6 +193,22 @@ independent of what the action declares.
 This is the reason the metadata must store the *terminal* resolved path rather
 than the immediate target.
 
+**(5) How a tree alias's contents are staged — symlinks are already content.**
+A tree alias never has to stage or relocate a symlink, because a tree artifact's
+value model contains none. A tree is a content-addressed, relocatable unit: its
+digest is a fingerprint over `(childPath, childContentDigest)` pairs, which is
+what lets any strategy materialize it at the alias's exec path (or a remote
+Merkle tree) from the digest alone. To preserve that,
+`TreeArtifactValue.visitTree` — the canonical tree capture — *dereferences* every
+on-disk symlink into the regular file or directory it points to (recursively)
+and *errors* on a dangling or tree-escaping link; the visitor is never handed a
+`SYMLINK` entry, so every child is a regular file with content and a digest.
+Consequently a tree alias re-addresses a child set that is already pure content:
+there is no symlink target to point in-sandbox vs. out-of-sandbox (the tension of
+part (1)), and parts (2)–(3) apply per child unchanged. This is *why* tree-to-tree
+aliasing is over regular files only — not a limitation of the alias, but a
+property of tree artifacts that the alias inherits for free.
+
 Illustrative (not normative) laydowns per strategy:
 
 | Strategy | `output` realized as | Bytes moved by Bazel |
--- a/designs/2026-07-20-artifact-alias.md
+++ b/designs/2026-07-20-artifact-alias.md
@@ -15,9 +15,10 @@ must present inputs at a canonical layout — rules_js co-locating sources under
 trees — express that need by *copying*, because copying is the only primitive
 available. This proposal adds artifact aliasing: `ctx.actions.alias(output,
 actual)` declares that `output` **is** `actual`'s content under a second exec
-path. `actual` may be a **file or a directory tree** — the same relation applies
-to both. It is not a copy and not a transformation; it is an identity relation on
-*content*. How that identity is realized on disk (symlink, hardlink, copy, or
+path. Both **file-to-file** and **tree-to-tree (directory)** aliases are in scope
+and supported; `output` and `actual` must be the same kind (file↔file,
+tree↔tree). It is not a copy and not a transformation; it is an identity relation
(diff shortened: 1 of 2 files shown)


designs/2026-07-20-artifact-alias.md
```diff
@@ -0,0 +1,137 @@
...
+downstream (aliasing-unaware) actions unchanged. One call per mapping.
+
+### Semantics: identity, not equality
+
+`output` and `actual` name the same content. This is an identity relation, with
+three consequences that distinguish it from a copy:
+
+- **They cannot diverge.** There is no moment at which `output` is "produced"
+  from a snapshot of `actual`; it is `actual`, re-addressed.
+- **The digest is shared, not coincidental.** `output`'s metadata delegates to
```
fmeum: This reads weird to me as part of the "Semantics" section. Shouldn't we specify the observable semantics from the point of view of a user and then derive ideas for transparent optimizations from these specified semantics?

designs/2026-07-20-artifact-alias.md
```diff
@@ -0,0 +1,137 @@
...
+### Laydown is the executor's choice, not the semantics
+
+Because an alias is an identity relation, no materialization mechanism is baked
+into it. Each strategy realizes it however it best presents the underlying
+content:
+
+- **Remote execution**: the Merkle tree places `actual`'s digest at `output`'s
+  exec path. The worker writes a real file; zero extra bytes. (This already
+  happens for `SymlinkAction` outputs.)
+- **Local / sandbox**: the strategy may symlink, hardlink, reflink/`clonefile`,
```
fmeum: I would like to read a more detailed explanation of what copies/reflinks/etc. where. If artifact A is declared as an alias of artifact B, then B is fully independent of A. If an action only has B as an input, how it is staged in the sandbox should matter more than how B is obtained from A in the actual exec root. 

Could you split this into separate parts?
1. How B is obtained from A in the physical exec root.
2. How B is expected to be staged by certain strategies when A is also an input.
3. How B I expected to be staged when A is not.

designs/2026-07-20-artifact-alias.md
```diff
@@ -0,0 +1,137 @@
...
+  symlink*, and the sandbox strategies have no handling of resolved paths at all
+  — which is why aliased content escapes the tree under Node's `realpath` and
+  rulesets fall back to copies. Aliasing makes a realpath-stable laydown
+  (content, not a followed symlink) available and cheap.
+
+Which laydown a strategy uses, and whether a given tool tolerates it, stays a
+configuration concern — Bazel makes correct-and-cheap laydowns *possible*, it
+does not promise any tool is realpath-safe. Practical constraints noted in the
+#396 thread carry over: default to copy or copy-on-write; hardlinks opt-in;
+never hardlink executables (macOS Gatekeeper).
```
fmeum: We need to exactly pin down the observable semantics or else Hyrum's law will force a particular strategy either way and the opt-in remains fictional. Hardlinks affect ctime and don't always work on MacOS, as you have noted

designs/2026-07-20-artifact-alias.md
thesayyn: Agreed, i did another pass on this, i think now its more honest about observable behavior today with different strategies.

designs/2026-07-20-artifact-alias.md
thesayyn: Added a section answering these questions, when symlink, reflink etc is used and different cases on how A, B, C is staged and another case where  A (actual) -> B (copy) -> C (copy) but both A and B is not part of the sandbox.

designs/2026-07-20-artifact-alias.md
thesayyn: Agreed, added those details. AFAIK most of this will just work and we can actually sidestep Gatekeeper issue by special casing executables for the Hardlinking execution strategy.   I saw most of these issues at https://github.com/bazelbuild/bazel/pull/29886  (AMFI, gatekeeper) etc so not my first rodeo.

lberki:
Why not link this from the main list in `README.md`? (put it in "Draft" state)

sluongng:
potentially duplicated of https://github.com/bazelbuild/proposals/pull/396/ ?

thesayyn:
> potentially duplicated of https://github.com/bazelbuild/proposals/pull/396

I described how this is different than what's on #396 and how it subsumes the proposal there.

thesayyn:
> Why not link this from the main list in README.md

Will do, i missed that.

lberki:
> > Why not link this from the main list in README.md
> 
> Will do, i missed that.

Far be it from me to be overly critical, but you haven't yet?

thesayyn:
Fixed, and fixed an earlier issue with sandbox proposal. thanks @lberki  

Don't mind me asking, what's the next step to get this to be accepted?

lberki (approved):

lberki:
We'll have to find a reviewer (and some capacity to review the change). Let me see what I can do.

fmeum:
Since this was sent as Markdown and not as a Google doc, should we comment on this PR or create a separate GitHub discussion? There is precedent for both :-)

Edit: Left some early general comments here: https://github.com/bazelbuild/proposals/pull/396#issuecomment-5061810133

